PR #29, opened 2026-07-19
feat(switch): silent switchdash session token refresh (CHOO-1435)

amaudruz:
## What & why

switchdash stores the gateway-issued `switch_auth` JWT (encrypted per-server) and replays it on every gateway call, but has **no refresh path** — once the JWT's 24h TTL elapses the next call 401s and the user is bounced to the sign-in screen. This adds a **silent session renewal** so an active session stays alive without user interaction.

Jira: [CHOO-1435](https://sandboxquantum.atlassian.net/browse/CHOO-1435)

## Approach (Option A — proactive sliding renewal)

The `switch_auth` cookie is a *stateless* HS256 JWT with no server-side session/refresh store, so "reactive refresh on a 401" isn't possible (an expired token can't authenticate its own renewal). Instead we renew **proactively, before expiry**:

**Gateway (switch-core)**
- New `POST /gateway/auth/refresh`, guarded by `get_current_user`, re-mints the `switch_auth` cookie from the still-valid session via the shared `set_session_cookie`.
- Provider-agnostic: re-issues from the `User`, so it renews **password and OIDC** sessions identically without replaying either login flow.
- An expired/invalid cookie 401s (a dead session cannot renew itself) → the client falls back to interactive sign-in. Fail-loud, no silent dead session.

**switchdash**
- `gateway-client.ts` renews opportunistically at **call time**: decodes the stored JWT `exp` and, when within 1h of expiry, refreshes via the endpoint before attaching the cookie. Evaluated per-call so it survives laptop sleep/wake.
- Concurrent authenticated calls are **deduped** to a single refresh round-trip (and one cookie write) per server.
- `refreshSession` in `auth.ts` persists the new cookie through the existing encrypted per-server store. Best-effort: network/rejection failures log a warning and fall through to the normal 401 → sign-in path rather than faking a live session.

## Scope notes
- No new persistent credential and no schema/migration — the endpoint renews against the existing session. A full refresh-token credential (survives long offline periods, enables reactive-on-401) was considered (Option B) and left as a potential follow-up gateway sub-task.
- The separately-tracked TTL bump (option B in the hub discussion) is out of scope here.

## Testing
- **Gateway:** `core/tests/switch_core/gateway/test_auth_refresh.py` — re-mints a fresh cookie for the current user; cookie is httponly and respects the `secure` flag.
- **switchdash:** `gateway-client.test.ts` — no renewal when far from expiry; renews + attaches fresh token when near expiry; falls back to current token when renewal fails; dedupes concurrent renewals; throws unauthorized without renewing when no token is stored.
- Local gate green: ruff + mypy + gateway pytest; switchdash typecheck + lint + format + vitest.

[CHOO-1435]: https://sandboxquantum.atlassian.net/browse/CHOO-1435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/switch_core/gateway/auth_routes.py
+++ b/core/switch_core/gateway/auth_routes.py
@@ -56,6 +56,30 @@ async def login(
     )
 
 
+@router.post("/auth/refresh")
+async def refresh(
+    response: Response,
+    user: Annotated[User, Depends(get_current_user)],
+    config: Annotated[SwitchConfig, Depends(get_config)],
+) -> UserResponse:
+    # Re-mint the switch_auth cookie from the still-valid session so an active
+    # client renews before expiry without re-authenticating — provider-agnostic
+    # (works for password and OIDC users alike, since it re-issues from the User
+    # rather than replaying either login flow). get_current_user rejects a
+    # missing/expired/invalid cookie with 401, so an expired session cannot renew
+    # itself; the client falls back to interactive sign-in in that case.
+    set_session_cookie(
+        response, user, config.jwt_secret_key, config.gateway_cookie_secure
+    )
+    return UserResponse(
+        id=user.id,
+        name=user.name,
+        email=user.email,
+        role=user.role,
+        created_at=str(user.created_at),
+    )
+
+
 @router.post("/auth/logout")
 async def logout(response: Response) -> dict[str, bool]:
     response.delete_cookie("switch_auth", path="/")
--- a/core/tests/switch_core/gateway/test_auth_refresh.py
+++ b/core/tests/switch_core/gateway/test_auth_refresh.py
@@ -1,0 +1,68 @@
+"""Tests for the silent session-renewal endpoint (CHOO-1435).
+
+`POST /gateway/auth/refresh` re-mints the `switch_auth` cookie for a caller
+whose current cookie is still valid, so an active switchdash client renews its
+session before expiry without bouncing the user to sign-in. The route depends
+on `get_current_user`, which already rejects a missing/expired/invalid cookie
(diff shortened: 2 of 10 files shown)


amaudruz:
## Update: expanded scope (per collaborating dev)

Merged latest `main` (the new switchdash-managed local Switch server) and added two follow-ons that share this ticket's "no forced re-login" intent:

**1. Managed local server is always signed in.** switchdash provisions the local server and holds its admin password, so `gateway-client` now resolves the auth cookie through a managed-aware path: when no session is stored, or a call returns 401, it silently re-logs-in with the stored admin secret (`reauthenticateManagedServer`) and retries once. The interactive sign-in is never shown for the managed server (first launch, reopen after the stack outlived the token, or mid-use expiry). Remote servers keep the proactive sliding refresh + fail-loud sign-in fallback.

**2. Authenticated gateway web pages.** The "open web app" (server view) and "open in gateway" (location) links now open an in-app Electron window with the server's `switch_auth` cookie injected into its isolated session, so the operator dashboard loads already signed in. The system browser can't be handed our httponly cookie, hence an in-app window. A same-origin guard keeps the cookie scoped to the gateway; the managed server mints a session first so the page is always authenticated.

New RPC: `switchServers.openGatewayPage({ serverId, url })`.

Tests: managed-server cases added to `gateway-client.test.ts`; new `gateway-web.test.ts`. Local gate green (typecheck, lint, format, vitest).

PR #79, opened 2026-07-27
feat(codex): Codex joins Switch via a connector plugin — drops the bespoke MCP setup path (CHOO-1436)

christian-mcdermott:
Part 2 of 2. **Targets #91**, not `main` — review that one first; this diff shows only the Codex work.

> **Reviewing note.** #91's branch predates #100, so GitHub currently renders this diff as 275 files — most of that is main's push-transport work arriving through the base. Against `main` the real change is **144 files**. Once #91 lands, the base retargets and this shrinks. `git diff origin/main...HEAD` is the honest view today.

Codex joins Switch the way every other agent does: a connector plugin, driven by the existing marketplace CLI machinery.

## Verified against the real Codex CLI (0.145.0 / 0.146.0)

Rather than reason from docs, this was built against an isolated `CODEX_HOME` and measured.

**Works:** Codex reads `.claude-plugin/marketplace.json` — the file this repo already ships — including Claude's **string** `source` form. One marketplace file serves both CLIs. The install CLI is non-invasive: a marketplace entry and `[plugins."x@y"] enabled = true`.

**Does not work, and this is what shaped the design:**
- **`${SWITCH_API_TOKEN}` is not expanded** in a plugin-bundled `.mcp.json`. A probe MCP server received the literal string `${SWITCH_API_TOKEN}`.
- **No `${CLAUDE_PLUGIN_ROOT}` equivalent**, so the plugin cannot reference its own install path.
- The `headers` block is silently dropped.

So the split is: **the plugin ships the skill**, and **switchdash registers the MCP server itself** when it launches the session.

## What's here

**`connectors/codex-plugin/`** — manifest + the Switch room-workflow skill, registered in the existing marketplace as `switch-connector-codex`. The skill is adapted, not copied. Codex rejects a manifest/marketplace name mismatch outright, so those are kept in sync.

**A `dialect` on the `cli` descriptor.** The review expected the existing marketplace driver to work unchanged. It doesn't — every field it parses is named differently (`pluginId`/`source.path` wrapped in `{installed, available}`; `plugin add`/`remove`; no per-plugin update verb; manifests under `.codex-plugin/`). `findInstalled` matched on `p.id === ref`, so Codex would have reported as **permanently not-installed with no error**. Both the local and remote (SSH) drivers now read a dialect table, with the verbatim JSON captured from Codex pinned in tests.

**Per-agent MCP registration via a Codex profile.** switchdash writes `$CODEX_HOME/<name>.config.toml` declaring the local `@sandbox-quantum/switch-agent-runtime` over stdio, and launches with `--profile <name>`. Measured details that are load-bearing:
- Codex hands an MCP child a **fixed allowlist** (`HOME`, `PATH`, `SHELL`, …), so the profile must name every variable it needs under `env_vars` — a server that names nothing starts with no credentials and dies before the handshake, reported only as `connection closed: initialize response`. Names only; no value ever reaches disk.
- A profile does **not** replace a same-named base-config server, it merges with it. A pre-profile `url` entry meets this one's `command` and Codex refuses to load the config *at all*, so switchdash removes such an entry first.
- `startup_timeout_sec` is raised, because a cold `npx` fetch from the private registry can exceed Codex's 10s default before the server does any work.

**Per-agent model / reasoning effort / instructions**, carried on the same profile.

**`switchSetup: kind: 'mcp'` is gone**, along with the Codex-specific auto-session prompt. With the skill installed and the hook registered, the generic path covers Codex.

## Changed by review: the room is connection-driven, not hook-scraped

An earlier revision tracked the room with a `connect_to_room` PostToolUse hook feeding the event enricher. **#100 replaced that**: switchdash opens the SSE connection, hands it to the session as `SWITCH_CONNECTION_ID`, and `connect_to_room` claims the room *on that connection*, so the server reports it back. Codex's PostToolUse hook is therefore **removed**, not added — `codex/hooks.test.ts` pins its absence.

Two blocking gaps that opening up created, both fixed here:

- **A UI-started remote session got no connection at all.** Only the local runtime and the sidecar's own spawner set `SWITCH_CONNECTION_ID`; the SSH path set nothing, the sidecar had no endpoint to mint one, and the hook that used to cover it was gone. The result was a Codex session that claimed its room on a connection nobody read, with `SWITCH_CHANNEL_DISABLE_POLL=1` discarding every frame — silence, where the identical Claude session worked. The sidecar now answers `POST /connection`; an older sidecar that 404s it **fails the start with an upgrade message** rather than coming up deaf, and a launch that dies after opening one hands it back so it cannot hold a room slot.
- **Remote sessions got no hooks.** `installRemoteHooks` served only `scope: 'workspace'` and Codex writes to `~/.codex`, so SessionStart never fired — every "resume" silently began a **new** conversation — and Stop never cleared the room's "working on it…". Global scope now installs through a shared remote `PluginFs` rooted at the VM home; the sidecar does the same for sessions it spawns; an unhandled scope raises.

## Two Codex defaults were justified by claims that turned out to be false

Both were measured against codex-cli 0.146.0 before being changed.

- **`sandbox_mode="danger-full-access"` is dropped.** It was justified by "every sandbox below full access blocks the loopback hook curls". **Codex runs hooks outside the sandbox** — a `SessionStart` hook curling `127.0.0.1` under `workspace-write` returns 200 / exit 0, while the same curl through `codex sandbox` is blocked. The loopback block applies to model-generated commands only. So the flag bought nothing and quietly removed containment "Bypass permissions" never offered to remove. Only `approval_policy="never"` remains.
- **Per-agent Instructions no longer wipe Codex's system prompt.** `model_instructions_file` **replaces** the model's instructions rather than adding to them: on the wire, a baseline `codex exec` sends `instructions` of 20,751 chars (Codex's whole operating manual); with the file set it sends only the file. The UI promised "Added to the agent". `developer_instructions` is additive (verified: 20,751 chars intact, body injected into the developer message) and — contrary to an early reading of `ConfigProfile` — **does work in the profile overlay**, which is also why `mcp_servers` works there. It is emitted there rather than on argv, which was measured to be both lossy and injectable on the tmux path.

## Other defects found and fixed

- **The generated hook command was broken shell.** `makeCodexStdinCommand` prefixed `printf '%s' "$INPUT" |` onto a string beginning with a `;`-separated preamble, so the port assignment ran in a subshell: the hook posted to `http://127.0.0.1:/hook` with an empty body, and `|| true` swallowed it. Pre-existing, and already installed in users' `~/.codex/hooks.json`. Replaced with the same `makeStdinHookCommand` Claude uses, and the test now executes each installed command under a real `sh` with `curl` stubbed rather than asserting on the command string.
- **Codex ran none of switchdash's hooks unless the agent auto-approved.** `--dangerously-bypass-hook-trust` was bundled inside the auto-approve flag; it is now a default arg. The tradeoff (per-invocation, also un-gates hooks the user added themselves) is recorded on the constant.
- **Remote auto-sessions could never launch.** Codex sets `sessionIdOnResumeOnly`, so a fresh session emits no id and the spec never contained `__SWITCHDASH_SESSION_ID__`, which `materializeAgentCommand` required — every spawn threw three times and posted "I tried to start a session but couldn't" into the room.
- **The storage migration could hand a non-Claude agent the Claude agent's token**, via an ungated `.claude/settings.local.json` fallback that applies to 30 of 31 providers. Recovered credentials are now behavior-gated *and* cross-checked against the agent row; a mismatch writes nothing and says so.
- **The #100 merge dropped tool-response shapes from the event enricher** (a bare content array, plus — found while restoring it — multi-block text and nested JSON strings). It is now genuinely the superset the merge commit claimed.
- **`cancel_task` was a phantom tool** — both skills and `protocol/instructions.py` told agents to call it; it was never registered. Exposed, with a test asserting every tool the skills advertise is registered.
- Skill accuracy fixes in both connectors (`list_participants` return shape, `read_visibility`/`write_visibility`, per-type `update_agent_detail`), and the Codex skill no longer claims the attachment tools don't exist — the profile registers the runtime that serves them.

## Known limitations, stated in code rather than papered over

- Codex versions only *installed* plugins, so **remote** update-detection has no advertised version to compare against. It returns "unknown"; the UI discloses that rather than reading it as up-to-date.
- No update verb → remove-then-add. A failed re-add leaves **no** connector, so that case gets its own message and its own test.
- `marketplace upgrade` only works on Git marketplaces, so update checks can't be exercised against a local checkout.
- Codex runs hooks via `$SHELL -lc`, so a `fish` or `nu` login shell gets a syntax error from the POSIX preamble.

## Release action before merge

`packages/switch-agent-runtime` is at **0.1.4**, but both pins (`.mcp.json`, `SWITCH_AGENT_RUNTIME_VERSION`) still name **0.1.3**, because a pin must name a *published* version. The `bin.ts` startup diagnostics therefore reach no session until `switch-agent-runtime-v0.1.4` is tagged and the pins follow.

## Verification

`format:check`, `lint`, `typecheck` clean across all five projects. **1689** desktop tests (node + main-db), **86** plugins, **162** core, **21** runtime, **16** migrations, app + sidecar bundles build. Python: `just check`/`typecheck` clean, **764** passed with 0 failures (the 145 errors are the documented no-Docker-socket testcontainer baseline). `browser` skipped per `dash/AGENTS.md`.


core/switch_core/gateway/known_agents.py
```diff
@@ -297,8 +297,123 @@ def start_session_instructions(
...
+    @field_validator("repo_dir", "notify_user", mode="before")
+    @classmethod
+    def _blank_string_to_none(cls, value: object) -> object:
+        if isinstance(value, str) and value.strip() == "":
+            return None
+        return value
+
+
+class CodexKnownAgent(KnownAgent):
+    connector_type = "Codex"
```
amaudruz: OpenAI Codex CLI ?

dash/docs/qa/codex-provider-CHOO-1436-test-plan.md
amaudruz, line 1: this should not be part of the repository (which will be open source at some point), that is internal team work process related stuff

amaudruz:
I am not sure about some of the statements made here, e.g.

`Codex has no Claude-style plugin-marketplace CLI, so it can't use Claude's connector`

I don't think that is true, after looking up some docs for codex in less than a minute I stumble into those
- https://learn.chatgpt.com/docs/skills-and-plugins
- https://learn.chatgpt.com/docs/plugins?surface=cli#build-your-own-plugin
- https://developers.openai.com/plugins/build/plugins

Which seems to indicate otherwise, so I would loop back and try the plugin approach instead that would be similar to Claude Code

dash/apps/switchdash-desktop/src/main/core/agents/known-agent-type.ts
```diff
@@ -0,0 +1,11 @@
+import type { AgentProviderId } from '@shared/core/providers/agent-provider-registry';
+
+/**
+ * Map a switchdash provider to the gateway known-agent type it registers as.
+ * Codex has its own gateway known-agent (correct connector type + `codex …`
+ * onboarding command instead of Claude's); every other provider still registers
+ * as `claude-code`, the generic switchdash-managed shape (CHOO-1436).
+ */
+export function knownAgentTypeForProvider(providerId: AgentProviderId): string {
```
amaudruz: error/warn if not claude or codex

dash/apps/switchdash-desktop/src/main/core/agents/register-agent-identity.ts
amaudruz, line 1: no defaults here, should be either codex or claude, (or error?)

dash/apps/switchdash-desktop/src/main/core/switch-rooms/auto-session-watcher.ts
```diff
@@ -40,6 +44,30 @@ const SPAWN_RETRY_DELAY_MS = 2000;
 // next notification retry.
 const INFLIGHT_TTL_MS = 120_000;
 
+/**
```
amaudruz: related to the commend I made in the PR discussion, codex should have ways to do plugins and skills -> if that is the case we should do something similart

dash/apps/switchdash-desktop/src/main/core/switch-servers/gateway-client.ts
```diff
@@ -218,13 +218,22 @@ export type RegisteredAgent = {
...
+     * inheriting Claude Code's (CHOO-1436). */
+    agentType?: string;
+  }
 ): Promise<RegisteredAgent> {
   const res = await gatewayFetch(server, '/agents/register', {
     authenticated: true,
     method: 'POST',
     body: {
-      agent_type: 'claude-code',
+      agent_type: params.agentType ?? 'claude-code',
```
amaudruz: again no default here

amaudruz:
```
The premise needs revisiting before this merges

The PR is built on "Codex has no Claude-style plugin marketplace CLI, so it can't use the connector." That's not correct, and it drives most of the design decisions here. Codex's plugin system is a near-clone of Claude Code's:

┌─────────────┬──────────────────────────────────────────┬─────────────────────────────────────────┐
│             │         Claude connector (ours)          │                  Codex                  │
├─────────────┼──────────────────────────────────────────┼─────────────────────────────────────────┤
│ Manifest    │ .claude-plugin/plugin.json               │ .codex-plugin/plugin.json               │
├─────────────┼──────────────────────────────────────────┼─────────────────────────────────────────┤
│ Skills      │ skills/*/SKILL.md                        │ same                                    │
├─────────────┼──────────────────────────────────────────┼─────────────────────────────────────────┤
│ Hooks       │ hooks/hooks.json                         │ same path, own event schema             │
├─────────────┼──────────────────────────────────────────┼─────────────────────────────────────────┤
│ MCP         │ .mcp.json                                │ same file name                          │
├─────────────┼──────────────────────────────────────────┼─────────────────────────────────────────┤
│ Marketplace │ claude plugin marketplace add owner/repo │ codex plugin marketplace add owner/repo │
└─────────────┴──────────────────────────────────────────┴─────────────────────────────────────────┘

It even reads legacy .claude-plugin/marketplace.json as a marketplace source, and supports local / git-subdir / npm entries. Manifest fields: name, version, description, skills, mcpServers, hooks, apps, interface.

So the way Codex should connect to Switch is the same way Claude Code does: a connector plugin. We should build one — connectors/codex-plugin/ alongside connectors/claude-code-plugin/, most likely sharing skills/ — rather than inventing a parallel kind: 'mcp' path inside switchdash. Codex then declares switchSetup: { kind: 'cli' } and the existing marketplace driver, the settings card, and the onboarding gate all work with no new capability variant.

Docs: plugins (https://learn.chatgpt.com/docs/plugins?surface=cli) · build plugins (https://developers.openai.com/plugins/build/plugins) · skills & plugins (https://learn.chatgpt.com/docs/skills-and-plugins)

Three things to verify first (an hour with a real Codex install), because they're the only real unknowns:

1. Does Codex expand ${SWITCH_API_ENDPOINT} / ${SWITCH_API_TOKEN} in a plugin-bundled .mcp.json? This is the PR's central justification for writing a literal URL into the global config. If it expands, the whole ensureSwitchMcpServerInstalled service is unnecessary.
2. Is there a ${CLAUDE_PLUGIN_ROOT} equivalenttdio server?
3. Install/update/uninstall CLI coverage (marketplace add/list/upgrade/remove are documented; install is desthe /plugins browser).---Hooks: Codex supports them, including tool-cal
Same issue. The PR works around the absence of a connect_to_room signal by pinning the poller to the spawn-tstartMcpRoomDelivery, once. If Codex later calfferent room — which it can; the tool is right there and list_linked_rooms invites it — the poller keeps delivering room A while the agent is in room B. Mesilently don't arrive, the sidebar nests the sm, and nothing warns. That's a fail-loudviolation.Codex supports PreToolUse, PostToolUse, Permismit, SessionStart, SessionEnd,SubagentStart/Stop, PreCompact/PostCompact. Pos MCP tools, receives tool_name / tool_input /tool_use_id, and filters by regex matcher. Our Codex plugin registers only three of those (Stop, PermissionRSessionStart) into the hooks.json it already w
                                                                                                            So we should use the same hook mechanism Claudd on the Switch connect_to_room tool, POSTingto the same 127.0.0.1:$SWITCHDASH_HOOK_PORT/hox hooks already use, routed throughagent-hook-service into setSessionRoom + switct. Identical path. startMcpRoomDelivery then
becomes a bootstrap that self-corrects on the _to_room.

Docs: hooks (https://learn.chatgpt.com/docs/hooks)

---
Codex is missing from Settings → Agents

CliAgentsList.tsx filters switchSetup.kind === main and 'mcp' after this PR — either way it's
filtered out, so you can't see or install Code tab at all. RECOMMENDED_IDS already contains
'codex'; it lands under Recommended the momenth a real Codex plugin (kind: 'cli') this needsno renderer change at all.

---
Onboarding gate is missing for mcp providers

if (kind === 'mcp') {
  onboardable.push({ agentId: plugin.metadata.id });   // unconditional
  continue;
}
if (kind !== 'cli') continue;
const status = await this.getStatus(plugin.met
if (status.installed) onboardable.push(...);

The cli branch is gated twice — binary resolveector plugin installed. The mcp branch is gated
zero times: Codex is offered in the add-agent o codex binary. We register the agentserver-side and write credentials to disk before discovering it can't spawn.

At minimum, gate on the host dependency (hostDady exists and the settings page derives status
=== 'available' from it). Better: hoist the binches. With the plugin in place, both gates
apply automatically.

The three readiness surfaces currently disagre

┌───────────────────┬────────────────────────────┐
│      Surface      │          Gate              │
├───────────────────┼────────────────────────────┤
│ Settings → Agents │ kind === 'cli'             │
├───────────────────┼────────────────────────────┤
│ Add-agent picker  │ kind === 'mcp' → always ry │
├───────────────────┼─────────────────────────┼────────────────────────────┤
│ Switch setup card │ kind === 'cli'             │
└───────────────────┴────────────────────────────┘

---
Credentials shouldn't be gated on repoAgents
                                                                                                                       .switch/agents/<slug>.json is switchdash-owned the only create-time writer wasbehavior.writeCredentials — a hook bundled into Claude's repoAgents capability. Codex has no repoAgents, so it fell    into the else and got nothing. This PR fixes tter into the else branch. It should instead beunconditional core behavior: switchdash alwaysy provider, and the plugin hook becomes
additive extras (Claude's permissions.allow, C

Related: Codex does have repo agent definitionproject-scoped, with a per-agent mcp_serverskey. But there's no --agent <name> equivalent  threads inside a session, not a launch-timesession identity), so repoAgents for Codex wouh is exactly why credentials must not hang offthat capability — it doesn't mean the same thi---Slug should be the agent name, not the agent iPost-CHOO-1440 the convention is name everywheded.name), remote watcher, reconciler, reset,sidecar controller (agent.name ?? agent.id), CThe one id-keyed reader is a leftover fallbackin local-agent-runtime.ts, and this PR builds it — plus fallback chains in two readers toguess between them. (This branch predates the ins it.) Fix: pass slug: params.name, fix thatreader, keep one key space.

---
Per-agent config: -c overrides rather than a global file write

ensureSwitchMcpServerInstalled writes [mcp_ser/config.toml, which is per machine — two Codexagents in one location share one entry and las

Codex's precedence is: -c/--config > project .ile > ~/.codex/config.toml > system > defaults. So -c is both the highest-precedence and the least invasive option: no global mutation, no collision, nothing stale to
clean up on delete, remote-safe (argv only), ags in buildCodexAutoApproveFlag. Worth
confirming -c accepts nested dotted keys (-c mif not, per-agent profiles
(~/.codex/<slug>.config.toml + --profile) are o the closest analog to Claude's --agent <name>
if we later want a launch-time identity select

---
Summary

The through-line across most of these: Codex iand worked around inside switchdash, when the
equivalent primitive exists and our plugin jus, skills, hooks, marketplace, repo agent
definitions — all present.

Concretely, I'd propose:

1. Spike the three plugin unknowns above.
2. Build connectors/codex-plugin/, declare Codli' }, and drop the kind: 'mcp' capability
variant — it's permanent public surface on theo gates (settings list, onboarding) have
already been missed by it.
3. Register PostToolUse for room tracking.
4. Make credential writing unconditional and name-keyed.
5. Gate onboarding on the host binary.

The CODEX_SANDBOX_MODE / CODEX_APPROVAL_POLICYmmand argv fix, and the test coverage are allgood and worth keeping regardless — those could land separately and quickly.

```

dash/docs/qa/codex-provider-CHOO-1436-test-plan.md
christian-mcdermott: Removed — the QA/test-plan doc is not in the branch tree or history anymore (dropped in the rewrite). `git ls-files | grep docs/qa` is empty.

dash/apps/switchdash-desktop/src/main/core/agents/known-agent-type.ts
christian-mcdermott: Addressed in cbf798e. `knownAgentTypeForProvider` now `log.warn`s and falls back to the generic `claude-code` shape for any provider outside the two server-side types, rather than mapping silently.

dash/apps/switchdash-desktop/src/main/core/agents/register-agent-identity.ts
christian-mcdermott: Addressed in cbf798e. `RegisterAgentInput.agentType` is now a required field (no default); the caller derives it via `knownAgentTypeForProvider`. An omitted type is a type error rather than a silent claude-code fallback.

dash/apps/switchdash-desktop/src/main/core/switch-servers/gateway-client.ts
christian-mcdermott: Addressed in cbf798e. `registerKnownAgent` takes `agentType` as a required param (no `?? 'claude-code'` default). The only hardcoded `agent_type: 'claude-code'` left is `registerKnownSubagents`, which is correct — subagents are a Claude-Code-only feature (Codex declares `repoAgents: { kind: 'none' }`).

dash/apps/switchdash-desktop/src/main/core/switch-rooms/auto-session-watcher.ts
christian-mcdermott: Done — this drove the full rewrite. Codex now joins via a connector plugin (`connectors/codex-plugin/`, `switchSetup: kind: 'cli'`, 1de30de), the bespoke `kind: 'mcp'` path and `startMcpRoomDelivery` are gone, and room tracking uses a PostToolUse hook on `connect_to_room` emitting `switch_room_connect` through the same enricher → `setSessionRoom` path Claude uses (c8e77a6) — so it self-corrects on every connect, including room hops. (The generated hook command also had a shell bug that made it a silent no-op locally; fixed in eb0eb05.)

core/switch_core/gateway/known_agents.py
christian-mcdermott: Using `Codex CLI`

christian-mcdermott:
## Refactored onto PR #100 (agent bridge push transport)

This branch has been updated to align the Codex provider with the local-stdio runtime + server-driven rooms that #100 landed, and to add per-agent Codex specialization. Three logical steps on top of the original work:

**Phase 0 — merge `origin/main` (#100)** (`02842a8`). Resolved conflicts favoring #100; ported the branch's `cancel_task` into #100's generic operation registry; kept the branch's logger-injected enricher.

**Phase 1 — Codex MCP via a per-agent profile** (`4c01225`, `a3a4390`).
- Register the local `@sandbox-quantum/switch-agent-runtime` (stdio) via a per-agent Codex **profile** (`~/.codex/<slug>.config.toml`, launched `--profile <slug>`) instead of the removed remote-HTTP-on-argv server. New pure `mcp.launchProfile` capability replaces `launchArgsForServer`. Written on all paths: local (home FS), SSH (`ctx.exec` → VM home), sidecar (baked into the launch spec as `launchFiles`).
- Dropped the Codex `switch_room_connect` PostToolUse hook — rooms are now driven by `SWITCH_CONNECTION_ID` (inherited provider-neutrally from #100).
- Removed dead argv-HTTP machinery (`switch-mcp-endpoint.ts`, the sidecar endpoint substitution, `assertSwitchMcpNameFree` — a profile cleanly overrides a user's own `switch` server, verified on codex-cli 0.146).
- Single source of truth for the runtime version pin, guarded by a test vs the Claude `.mcp.json`; sidecar version bumped.

**Phase 2 — per-agent model / effort / instructions** (`7fb3f50`).
- The same profile now carries `model` + `model_reasoning_effort`, and (when set) a `model_instructions_file` written as a second file. Stored in a versioned JSON column `agents.provider_config` (migration `0045`, additive nullable). Loaded at launch on all three paths. New "Codex configuration" section in the add-agent modal.

Every Codex profile behavior was verified empirically against the installed **codex-cli 0.146.0**. Green: typecheck, lint, format, 1579 app + 162 core + 73 plugin + 16 migration tests.

⚠️ The original PR title/description predate this refactor.

PR #129, opened 2026-08-05
feat(codex): standalone setup via a configure skill — connect to Switch without switchdash (CHOO-1936)

christian-mcdermott:
Ships a `configure` skill in the Codex connector plugin so a user can connect
Codex to a Switch server with the **plugin alone** — install, run the skill,
and `codex` reaches Switch with no switchdash involved.

Jira: [CHOO-1936](https://sandboxquantum.atlassian.net/browse/CHOO-1936)

## What it does

Registers the agent against the bridge (`POST /agents/register-known`,
`agent_type: "codex"`), then registers the `switch` MCP server with
`codex mcp add`, storing credentials as **literal values** in the user's base
config.

Literal values rather than the `env_vars` name-forwarding the switchdash
profile uses, because the two paths differ in who launches Codex. `env_vars`
forwards names from the launching process's environment — which works
precisely because switchdash *is* that process and populates it. Standalone
there is no injector, so names would forward nothing and the runtime would die
before the MCP handshake.

## Behaviour verified against codex-cli 0.146.0

Rather than assuming parity with Claude Code, the design was driven by testing
against a throwaway `CODEX_HOME`. Three findings shaped it:

- **An `mcp_servers.<name>` table must declare its own transport.** An
  `env`-only table — the obvious way to add credentials to a server the plugin
  already registers — is rejected with `Error loading config.toml: invalid
  transport`, and that failure takes down **every** Codex session on the
  machine, not just the Switch tools. Hence the skill drives `codex mcp add`
  (which always writes a complete entry) rather than editing TOML, making the
  trap structurally unreachable.
- **Codex has no project-scoped config.** `./.codex/config.toml` is not read
  for `mcp_servers`; config is `$CODEX_HOME` only. So the Claude connector's
  per-project/global scope choice has no analogue — the standalone identity is
  per-machine.
- **The offline run command Switch posts never includes `--profile`** (it is a
  bare `cd "<repo_dir>" && codex "connect to switch room <name>"`), so base
  config is the only placement where that paste-ready command works as
  written.

## Adapted, not copied

Two steps from the Claude Code configure skill are **dropped**, with the
reason stated inline so they don't get "restored" later:

- The channels question — `CodexOptions` has no `channels_enabled` field;
  sending one is silently ignored, and Codex can never be `session_passive`.
- Subagent registration — Codex has no `.claude/agents/*.md` equivalent, and
  `register-known-bulk` would happily mint a batch of indistinguishable
  children.

`auto_session` is deliberately left `false`: it means "switchdash watches
rooms and auto-spawns a session", which standalone has nothing to honour.

## Runtime pin

Read from the installed plugin's `.mcp.json` rather than hardcoded — a
hardcoded number would be a **third** pin, outside the parity test guarding
the other two, and would go stale silently. If absent, the skill stops and
asks for a plugin upgrade rather than inventing a version. Uses `find` rather
than a glob, since an unmatched glob is a hard error in zsh (macOS default).

## Documented sharp edge

A base entry's literal `env` **wins** over a profile's `env_vars` forwarding,
per key. So running this skill on a machine that also uses switchdash makes
switchdash's sessions authenticate as the standalone agent — wrong identity,
no error. The skill checks for `~/.codex/*.config.toml` and warns before
writing; the README documents it.

Related: the codex-plugin README claims switchdash "removes such an entry when
it writes the profile". It does not — `codexMcpAdapter.removeServer` has no
call sites. Noted in the README rather than fixed here, since that is
switchdash code and belongs with CHOO-1935.

## Sequencing

Independent of CHOO-1935 — the skill must write a transport-bearing entry
either way, so it works with or without the bundled `.mcp.json`, and the two
can merge in either order. Note both branches bump `plugin.json` to `0.2.0`;
whichever merges second takes a one-line conflict and should go to `0.3.0`.

## Verification

End-to-end verification against a live Switch server (install → configure →
registered → connect → post), in particular what actually happens with inbound
event delivery standalone.

[CHOO-1936]: https://sandboxquantum.atlassian.net/browse/CHOO-1936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amaudruz:
**The plugin already ships the MCP server — this adds a second one with the same name.**

`connectors/codex-plugin/.mcp.json` (from CHOO-1935, now on `main`) declares `mcp_servers.switch`: the runtime, `startup_timeout_sec: 60`, `default_tools_approval_mode: "approve"`. Step 7 then runs `codex mcp add switch ...`, writing a second complete `switch` entry into base config. The Sequencing note says the skill works "with or without the bundled `.mcp.json`" — that was fair while both were in flight, but "without" is no longer a case.

**The gap it's filling is real.** The plugin supplies the server *definition*, not the *identity*: `env_vars` only forwards names, and standalone nothing sets them, so the runtime exits at `bin.ts:85` before the handshake. Something must supply credentials. The question is whether cloning the whole server entry is the smallest way.

**Suggested direction: let the runtime resolve its own credentials.**

Today `bin.ts` reads three env vars and hard-exits if any are missing. Make that a chain instead:

1. **Env vars, if set** — switchdash's path, unchanged and unaffected.
2. **Otherwise, a local `.switch/agents/` store** — if exactly one agent, use it.
3. **If ambiguous, ask.** Expose one extra `select_agent` tool; calls to anything else return a plain string naming the available agents. The model picks, and the identity binds for the session.

The tool catalog stays static and complete (bootstrap the `/ops` fetch off any local entry — same server, same catalog), so this needs no dynamic tool-list support from Codex. One guard at the `CallToolRequestSchema` chokepoint, not 50.

What that buys:

- The plugin's `.mcp.json` stays the **single** server definition — no duplicate entry, no name collision, approval mode and version pin preserved.
- **This skill shrinks a lot**: register the agent, write a credentials file. The version-pin scraping, the hand-patched `startup_timeout_sec` and the `codex mcp add` plumbing all go.
- **Fixes the Claude connector too** — one store replaces its `${VAR}` expansion in `settings.local.json`.
- "One identity per machine" stops being a permanent caveat.

Two things to settle:

- **Where the token lives.** Project-local selection is nice, but a live API token inside a git repo is a hazard gitignore can't reliably contain. Option: keep the *name* in `./.switch/agents/`, the *secret* in `~/.switch/agents/` at 0600.
- **Publish lag.** This is a `bin.ts` change, so per `dash/AGENTS.md` the pins can't move until `switch-agent-runtime-v<version>` is tagged. Real cost versus the current approach; worth stating up front.

Also: both this and CHOO-1935 bump `plugin.json` to `0.2.0`, so this needs `0.3.0`.

christian-mcdermott:
Agreed on all counts, and I've measured the config-merge behaviour rather than reasoned about it. **It is worse than a second entry with the same name: it is a clobber.**

Tested against the real post-1935 plugin (codex-cli 0.146.0, isolated `CODEX_HOME`), `codex mcp get switch` before and after Step 7's `codex mcp add`:

```
plugin only                          after `codex mcp add switch ...`
  startup_timeout_sec: 60              (absent)
  default_tools_approval_mode: approve (absent)
```

A base-config entry does not merge per-key with a plugin-provided server — it **replaces** the definition. So Step 7 silently strips both the approval mode and the timeout. Given `approve` is the only value that admits a write-annotated tool, this PR as it stands would leave a standalone agent unable to `post_message` in exactly the unattended case the mode exists for. That is a regression, not just redundancy, and it settles the question of whether cloning the entry is acceptable: it isn't.

The `.mcp.json` staying the single definition is therefore not only tidier, it is the only shape that preserves those two properties.

**Your direction checks out against the code.** `bin.ts:85` is a single hard-exit on the three env vars, so the resolution chain slots in without touching the startup path elsewhere; and there is exactly one `CallToolRequestSchema` handler (`bin.ts:585`), so the guard really is one chokepoint rather than fifty.

**One gap worth settling before implementation.** The catalog bootstrap is `GET {API_ENDPOINT}/agents/{AGENT_ID}/ops` — agent-scoped, not server-scoped. "Bootstrap off any local entry — same server, same catalog" holds only while every entry in the store points at one Switch server. With entries across two deployments, picking an arbitrary entry can bootstrap the catalog from the wrong one, and the mismatch would surface as tools that exist but 404 on call. Either group the store by endpoint and bootstrap within a group, or defer the catalog fetch until after `select_agent` binds an identity.

**On your two open questions:**

- **Token location** — agreed, and for the reason you give: name in `./.switch/agents/`, secret in `~/.switch/agents/` at `0600`. A live API token inside a working tree is not something gitignore reliably contains, and the failure is silent and permanent once pushed.
- **Publish lag** — accepted and worth stating plainly, but it does not block verification: the runtime can be exercised locally by pointing the MCP command at the built path instead of `npx`, so the change can be proven before the tag exists and the pins move.

**Sequencing.** Agreed with christian.mcdermott to split this: the credential store is a change to the published runtime, it fixes the Claude connector's `${VAR}` expansion in `settings.local.json` too, and it is not really CHOO-1936-shaped. A separate ticket will carry it, and this PR waits on it — I am marking it draft rather than leaving a mergeable-looking PR with a known clobber in it. Once the runtime resolves its own credentials, this skill loses Step 7, the pin scraping and the timeout patch, and becomes: register the agent, write the credentials file.

Plugin bumped to `0.3.0` and `origin/main` merged in, as you noted.

amaudruz (approved):

christian-mcdermott:
**Rewritten onto the runtime credential store (CHOO-1962). The mechanism this review rejected is gone.**

Merged `origin/main`, and the skill no longer touches MCP config at all. `codex mcp add`, the runtime-pin scraping and the hand-patched `startup_timeout_sec` are all deleted. It now registers the agent and writes `.switch/agents/<name>.json` in the working directory — the plugin's `.mcp.json` stays the single server definition, so `default_tools_approval_mode: approve` and the timeout survive.

The skill states the clobber explicitly, so the entry does not get reintroduced by someone solving the same problem again.

**Verified against a Switch server built from this branch, with the 0.2.0 runtime:**

- The skill registered `codex.choo1936.standalone` and wrote the store at mode `0600`, with `.switch/agents/.gitignore` created *before* the token — confirmed the file is genuinely ignored, not merely untracked.
- A session with **no** `SWITCH_*` environment, started in that directory, resolved that identity from the store, opened its stream and served 47 tools with a live `list_rooms`:
  ```
  switch: running (agent_id=37e94a6f-ad90-460f-a02a-9cac5ee24b73)
  switch: connection established (rooms=[])
  ```
- Two agents on one server → `running unbound — 2 agents provisioned here, awaiting select_agent`, and every other call refused while naming both.
- Two agents on **different** servers → refuses at startup with the disambiguation message. That was the gap I raised on the original comment; confirmed fixed and behaving.

The last two were driven directly over stdio rather than through a model, so they test the runtime contract rather than a model's interpretation of it.

**One new hazard, now documented in Step 1.** A complete `SWITCH_*` environment silently takes precedence over the store, so a shell that exports one — a switchdash-spawned terminal does — makes everything this skill writes inert, with no warning. The skill now checks `printenv` for all three and says so before writing.

---

⚠️ **Merge consideration: the feature is not reachable by users yet.** The credential store ships in runtime **0.2.0**, which is **unpublished** — the registry's latest is `0.1.6`, and all three pins (both connectors and switchdash) still name `0.1.6`. There is no `switch-agent-runtime-v0.2.0` tag on origin. Until that tag is pushed and the pins move, a user who runs this skill gets a runtime that cannot read what it wrote. My verification used a local build to get around exactly that.

So this is correct and reviewable now, but it only becomes *true for users* once 0.2.0 is published. Worth deciding whether it merges before or after that.

---

Unrelated, and flagging rather than touching: this PR's parent (CHOO-1962) is titled *"the token leaves the working tree"* and its commit messages describe a `$HOME`-scoped store at `0600` keyed by agent id. **That did not land.** The merged `credentials.ts` reads the project directory only, has no `homedir` reference, and its own docblock says the split was deliberately deferred. `write-switch-settings.ts` also carries a comment asserting *"since CHOO-1962 a provisioned agent's token lives under `$HOME`"*, which is now false. The code is self-consistent — token in the working tree, git-ignored — but the title claims a security property the code does not have, and someone will rely on it.

christian-mcdermott:
**0.2.0 is published and tagged, so the pins have moved. The standalone path is now real rather than documented, and verified against the published artifact.**

`switch-agent-runtime-v0.2.0` exists on origin and `0.2.0` is on the registry, so per AGENTS.md the three pins that name what sessions actually run can follow. All three move together because the parity test requires it — both connectors must register the same published version as `SWITCH_AGENT_RUNTIME_VERSION`, and no pin may run ahead of `package.json`. Claude plugin bumped to `0.8.1` since its `.mcp.json` changed. The suite passes (10/10).

**This also repairs a transient breakage on `main`.** CHOO-1962 removed the credential block from the Claude connector's `.mcp.json` on the grounds that the runtime resolves its own identity — but the pins still named `0.1.6`, which cannot read the store. So on `main` today a standalone session on *either* host has no identity at all: Claude declares no credentials and can't fall back, and Codex forwards names nothing sets. Moving the pins closes that.

**Verified with the published package** — no local build, no scaffolding:

```
$ npx -y @sandbox-quantum/switch-agent-runtime@0.2.0     # cwd = project, no SWITCH_* in env
switch: running (agent_id=545651ab-c085-4585-8231-44e84b2e8366)
switch: stream open (connection 06fd61bc-…, cursor 0)

create_room -> {"id":"b38baf05-62c5-416c-958c-e408759e8029",
                "matrix_room_id":"!a4QNucsrVdTJLJzl5n:localhost"}
```

Register an agent, write `.switch/agents/<name>.json`, and a session with no environment at all resolves that identity and creates a real room through it. That is the ticket's acceptance criterion met with the artifact users will actually run.

One observation, not a blocker: the first heartbeat came back `404 — reopening stream`, and the runtime recovered on its own and worked normally afterwards. It looks like a benign race on a freshly-opened connection rather than a fault in this path, but it is reproducible enough to be worth someone's eye if it shows up elsewhere.

christian-mcdermott:
## Review — request changes, do not merge in current state

I authored this, so I reviewed it looking for reasons it is wrong rather than reasons it is fine. Two of the findings break things well beyond this ticket.

### Blockers

**1. It does not merge.** Five conflicting files: both connectors' `.mcp.json`, `codex-plugin/.codex-plugin/plugin.json`, `console/AGENTS.md`, and `switch-agent-runtime.ts`. A dry-run merge of `origin/main` confirms it.

**2. It would point every session at a registry the project has left.** Main moved the runtime to a new scope on the public registry — `@sandboxaq/switch-agent-runtime@0.3.0` (`6e03956d`, "first npmjs publish under @sandboxaq"). This branch still carries `@sandbox-quantum/switch-agent-runtime@0.2.0` in **both** connectors, so merging rewrites Claude's and Codex's `.mcp.json` backwards: wrong scope, older version, private registry. Sessions on both hosts would fail to fetch the runtime unless the user still holds GitHub Packages auth. A Codex-documentation PR breaking the Claude connector is the worst version of this.

**3. CI is red.** Reproduced locally:

```
test_the_registry_agrees_with_every_file_it_records
  switch-connector-codex: artifacts.yaml says 0.2.2, plugin.json says 0.3.0
  switch-connector:       artifacts.yaml says 0.8.0, plugin.json says 0.8.1
```

Two plugin bumps without the matching `artifacts.yaml` update — precisely the drift that test exists to catch. My error, and I compounded it by reporting "tests green" from local runs without checking the pipeline.

**4. Version collision.** Main released the Claude plugin as `0.8.1` (`1dc74182`). This branch also claims `0.8.1` with different content.

**5. It modifies code main deleted.** `SWITCH_AGENT_RUNTIME_VERSION` and its tests were dropped in `62e9ed42`. This branch still edits that constant.

**6. Roughly 76 lines of the skill are now actively harmful.** The repo is public and the private-registry machinery is gone — `npm.pkg.github.com` has zero references left on main, and `switch-agent-runtime-publish.yml` targets public npmjs "rather than to a registry that demands a credential to read." Step 0 nevertheless still walks the user through `gh auth refresh -h github.com -s read:packages` and writing a **token into `~/.npmrc`**, i.e. creating a credential at rest for no reason. Main's Claude configure skill has already dropped its equivalent step; this reintroduces the pattern. The troubleshooting section compounds it by diagnosing any 404 as registry auth.

### Should fix

- The retired `@sandbox-quantum` scope is named 8 times across the skill and README.
- The README's "npm environment is part of the contract" section is built on the private-registry premise. The auth-specific half should go; the transferable part (Codex strips `npm_config_*`, so verify the way the server runs, not the way your shell does) is worth two lines.
- **The approval on this PR is stale** — submitted 2026-08-10T13:58Z, before the branch's final commit at 19:11Z and before every structural change on main. It should not be read as sign-off on the current diff.

### What holds up

The architecture is unaffected. I re-checked the foundation rather than assuming it: `credentials.ts` at `0.3.0` still reads `.switch/agents/` from the working directory, with the same store shape and resolution order as `0.2.0`. So the substance — register the agent, write the store, touch no MCP config — plus the clobber warning, the working-directory semantics, `select_agent`, and the multi-server refusal all remain correct and verified. Every defect above is staleness or packaging.

### Remediation

Merge main taking **main's** side for both `.mcp.json` files and the deleted constant; Claude plugin to `0.8.2`; update `artifacts.yaml` for both plugins; delete Step 0 and its troubleshooting entries; `@sandbox-quantum` → `@sandboxaq` throughout; re-run CI; re-verify end to end against `0.3.0`.

One judgement call worth a second opinion before I do it: drop Step 0 **entirely**, matching main's Claude skill, or keep a short note that `npx` must be able to reach the public registry?

amaudruz (approved):

PR #137, opened 2026-08-06
fix(switchdash): report a missing remote working directory instead of failing at write time (CHOO-1416)

amaudruz:
## Problem

Creating a Switch agent on a remote host hard-failed when the chosen working directory's parent didn't exist:

```
Failed to register agent
Error: Error invoking remote method '...': FileSystemError: File or directory not found: /home/ubuntu/switch-agents/internal-deployments
```

[CHOO-1416](https://sandboxquantum.atlassian.net/browse/CHOO-1416) · P0

## Root cause

`SshFileSystem` is rooted at the agent's working directory, and `ensureRemoteDir` walks up creating missing parents but refuses to go above its own root:

```js
parentPath.length >= this.remotePath.length
```

The comparison is `>=`, so the walk reaches — and creates — the working directory itself, but nothing above it. A missing leaf under an existing parent has therefore always worked; a missing *parent* is what cannot be recovered.

So for `remoteRepoDir = /home/ubuntu/switch-agents/internal-deployments` with no `/home/ubuntu/switch-agents`, the first credentials write (`.switch/agents/.gitignore`) walks `…/.switch/agents` → `…/.switch` → the repo dir, then rejects.

Two consequences:

1. The user gets a raw `FileSystemError` for what is the only unvalidated input in the flow — the SSH host is probed for reachability and the server is picked from a list, but the working directory is free text nothing checks.
2. It fails **after** `addAgent` has minted an identity on the gateway, leaving an agent that exists in Switch and nowhere else.

The ticket names `switchServers.provisionRemoteAgent`; the live path is now `agents.addAgent`. Same bug, and `provisionRemoteAgent` has had no caller since CHOO-1440 (see below).

## Approach

The containment guard is a sandbox boundary — an fs rooted at a working directory should not be able to create that directory's ancestors — so it is unchanged. Instead the directory is checked before anything is minted, and the unrecoverable case is reported rather than worked around.

`inspectRemoteDir` (`main/core/agents/remote-dir.ts`) answers it in two stats, from an fs opened at the host root since one rooted at a missing directory cannot stat its way out to look at the parent:

| status | meaning | outcome |
| --- | --- | --- |
| `directory` | exists | proceed |
| `creatable` | missing, parent exists | proceed — the write creates it, as it always did |
| `file` | exists, but is a file | refuse |
| `missing` | parent is missing too | refuse |

`isUsableRemoteDir` keeps the UI gate and the server-side guard agreeing on which of those are acceptable.

- **The add-agent modal** probes when the working directory is committed ("Set location") and shows a one-line notice for the refused cases, modelled on the adjacent `HostReachabilityNotice`. Submit is gated on a usable directory. `creatable` shows nothing.
- **`addAgent`** performs the same check before `registerAgentIdentity` and returns a typed `directory-missing`, so the guarantee doesn't depend on the UI gate.

### What this deliberately does not do

switchdash does not create the directory for you. Making one on someone's host is their decision, and a hand-typed path is exactly where "helpfully" running `mkdir -p` turns a typo into a real, wrong directory with an agent in it. Where the existing write already handles it (`creatable`) nothing changes; where it doesn't, the flow says so.

Probing before the mint also means this failure no longer orphans a Switch-side agent. That is re-ordering, not atomicity — general rollback remains [CHOO-1415](https://sandboxquantum.atlassian.net/browse/CHOO-1415), which this does not close.

### Drive-by removal

`switchServers.provisionAgent` / `provisionRemoteAgent` lost their last renderer caller in `d6a34b8` (CHOO-1440, flat agents) and wrote the superseded `.claude/settings.local.json` layout. Removed, along with `ProvisionAgentParams` / `ProvisionRemoteAgentParams` / `AgentProviderKind`, which nothing else used. `ProvisionAgentResult` stays — `register-agent-identity` and the modal still use it.

## Testing

`pnpm typecheck`, `pnpm lint`, and the `node` + `main-db` vitest projects: **1826 tests passing**.

- `ssh-fs.test.ts` pins the containment behaviour using the ticket's exact path — the walk-up sequence and the exact error string — so it isn't "fixed" later by widening the guard, plus the converse case rooted at `/`.
- `remote-dir.test.ts` covers all four statuses, including `creatable` explicitly so the pre-existing leaf-creation path can't be gated away again, parent-is-a-file, a non-absence probe failure propagating, relative-path rejection, trailing-slash normalisation, and SFTP channel cleanup on the throw path.
- `add-agent.test.ts` covers the refused cases returning `directory-missing` **with no identity minted**, `creatable` proceeding to creation, the happy path, and no SSH probe for local agents.

Manually verified on a live remote host by @louis-amaudruz during review.

## Reviewer notes

- Importing `remote-dir` into `add-agent.ts` pulls `ssh-fs` → the Electron-bound db into that module graph at load, which broke `add-agent.test.ts` until mocked. Worth knowing before importing anything SSH-shaped there.
- `writeSwitchSettings` / `writeRemoteSwitchSettings` are now dead (the removed handlers were their last callers) but are left in place to keep this PR about the bug. Happy to remove them here or in a follow-up.

[CHOO-1416]: https://sandboxquantum.atlassian.net/browse/CHOO-1416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHOO-1415]: https://sandboxquantum.atlassian.net/browse/CHOO-1415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abeldantas:
Superseded by #373: the fix (inspectRemoteDir preflight + typed refusal before minting) is ported to current main there, fix-only. This branch predates the dash-to-console rename and its dead-code sweep clashes with how main evolved (some deleted files are load-bearing again), so the sweep is worth a fresh audit on today's tree if still wanted. Closing to keep WIP honest.

PR #141, opened 2026-08-06
feat(switch): runtime resolves its own Switch credentials, and the token leaves the working tree (CHOO-1962)

christian-mcdermott:
A connector plugin supplies the Switch MCP server's *definition* but not the
*identity*, so a standalone session had nothing to be and the runtime exited at
`bin.ts:85` before the MCP handshake. This makes the runtime resolve its own
credentials, and moves the API token out of the working tree while it is there.

Jira: [CHOO-1962](https://sandboxquantum.atlassian.net/browse/CHOO-1962) ·
unblocks [CHOO-1936](https://sandboxquantum.atlassian.net/browse/CHOO-1936)

## Why not just add a second `mcp_servers.switch` entry

Measured against codex-cli 0.146.0: a base-config entry **replaces** the
plugin's definition wholesale rather than merging per key, silently dropping
`startup_timeout_sec: 60` and `default_tools_approval_mode: approve`. Losing
`approve` means write-annotated MCP tools are refused — so a standalone agent
cannot `post_message`, in exactly the unattended case the feature exists for.
Not a duplicate-server annoyance; a clobber that breaks the agent quietly.

## The resolution chain

1. **Env vars, taken whole when all three are present.** switchdash injects them
   per session and accounts for most sessions; that path is untouched.
2. An unexpanded `${SWITCH_*}` counts as **absent**, which is what lets a Claude
   connector session fall through instead of dying on its own placeholder.
3. **Half an environment refuses to start.** Falling through on a partial env
   could authenticate as a different agent than intended.
4. **Store, one server, one agent** → bind it.
5. **Store, one server, several agents** → identity left open, `select_agent`
   binds it, one guard on the call handler refuses everything else and names the
   options. Stream and heartbeat start **on bind**, not at boot.
6. **Store, several servers** → `failStartup` naming the conflict;
   `SWITCH_API_ENDPOINT` alone disambiguates.

### On the "agent-scoped bootstrap" gap

The ticket flagged `GET /agents/{AGENT_ID}/ops` as agent-scoped, so bootstrapping
off an arbitrary local entry could hit the wrong server. It is agent-scoped for
**authentication only** — `get_operations` never uses `agent_id` or the resolved
`agent`, and `list_operations()` takes no agent and answers from the static
registry. The catalog is a property of the **server**, not the identity.

So the gap is narrower than stated: it only bites when entries span different
**endpoints**. Grouping by endpoint addresses it exactly, and any candidate's
credentials bootstrap the catalog correctly (they are tried in turn, so one
revoked token does not masquerade as an unreachable Switch). Multi-endpoint
fails loud rather than serving a `select_agent`-only catalog, which would have
made the catalog dynamic — contradicting the ticket's "the tool catalog stays
static" and betting on hosts honouring `tools/list_changed`.

`select_agent` stays listed after binding for the same reason.

## The token split

`.switch/agents/<slug>.json` keeps its shape and loses one key. Endpoint and
agent id stay — neither is a secret, and **the file doubles as the settings file
Claude Code is launched with** (`subagents.ts:460`), so its shape is a contract
beyond switchdash. Only `SWITCH_API_TOKEN` moves, to
`$HOME/.switch/agents/<agent-id>.json` at `0600`, on whichever host the agent
runs on.

- **Keyed by agent id, not name.** `$HOME` is shared by every project, so a
  name-keyed secret would let two agents called `reviewer` overwrite each
  other's token — the CHOO-1960 collision with a wider blast radius. No conflict
  with that ticket; it stays free to solve project-side naming.
- **`PluginFs` cannot set a mode** — five methods, six implementations, and
  `FileSystemProvider` exposes `mode` only for reading. Widening it would mean
  changing every implementation and test double for one caller, so this follows
  `ServerHost.writeFile` / `npm-registry-auth` instead: one small store per
  transport. Remote writes set `umask 077` *before* the redirect, so the file is
  never briefly world-readable with a live token in it.
- **Migration generation 3** relocates tokens already on disk. Generations 1 and
  2 only wrote the neutral file when it was *absent*, so without this the token
  would move for new agents and nobody else. It also ends the generation-1 skip
  for behavior providers — the token can be in any agent's file, and only the
  file can say.
- **`deleteAgent` removes the stored token.** Deleting the working-tree file
  removed only the pointer, which would have stranded a live credential.

## Sidecar: 1.8, not 2.0

By the letter of `sidecar-version.ts`, changing a shared on-disk layout is a
major. But a major only achieves anything if `MIN_SUPPORTED_SIDECAR_MAJOR` moves
with it, and that kills every older sidecar on sight — including one an older
switchdash on the same host kills right back, forever (CHOO-1937). The sidecar
reads **both shapes** instead, so the question does not arise.

## Publish lag

`package.json` goes to **0.2.0**; the two pins stay at **0.1.5**. `AGENTS.md`
requires them to name a *published* version, so they move only after
`switch-agent-runtime-v0.2.0` is tagged. **Until then this `bin.ts` change
reaches no session.** For verification, point the MCP command at the built path
instead of `npx`.

## `.claude/settings.local.json` too

The same exposure in the other working-tree file, and it could not be fixed
alone. That file's `env` block is what expands `${SWITCH_API_TOKEN}` in the
Claude connector's bundled `.mcp.json`, and a `${VAR}` resolving to nothing does
not degrade — it **fails the whole MCP server**. So:

- The connector declares **no `env` block at all**. Nothing is lost: switchdash
  sets real environment variables, which reach an MCP child regardless, and a
  standalone session resolves from the store. Declaring them would break exactly
  the case the store exists for — the drift test now asserts their *absence*.
- `mergeSwitchSettings` writes no token and strips one it finds, making it
  identical to the neutral merge, so the two collapse into one function.
- `mergeSwitchApiEndpoint` no longer needs a token to recognise a provisioned
  agent. It did — so after the migration a server-URL change would have silently
  skipped every migrated agent and left it on the old endpoint.
- Provisioning writes the neutral store as well, or an agent provisioned that
  way would have a token nowhere at all.
- The `configure` skill writes both halves, and no longer detects
  "already configured" by looking for a token.

## Live verification

Run twice — before and after the `settings.local.json` change — against the
running `switchdash-dev-local` Switch on `127.0.0.1:52917`, driving the **built**
`dist/bin.mjs` over stdio with two agents genuinely registered through
`POST /agents/register-known`. **17/17 both times.**

Covered: single-agent bind with no env vars; the real **46-operation** catalog
fetched from the live server; an authenticated `list_rooms`; unbound start with
two agents; `select_agent` offered and everything else refused with both
candidates named; binding and then calling a real tool; env vars still winning
over an ambiguous store; multi-endpoint refusing to start and saying nothing on
stdout; the secret at `0600`.

Caveat: that instance is an older build knowing only `claude-code`, so both
agents were registered as that. The resolution path is agent-type agnostic, but
a *Codex* session has not been exercised end-to-end here.

## Testing

switchdash **1797** node + **49** db/migration · plugins **89** · core **162** ·
runtime **38**. Typecheck, lint and format clean; `pnpm build` succeeds and the
sidecar bundle carries the store with no Electron leak.

New coverage for the token-less case on every path touched: store-only startup,
`${VAR}` fall-through, multi-endpoint refusal, partial-env refusal,
`select_agent` offered/refusing/binding, generation-3 relocation, delete removes
the secret (and leaves another agent's alone), preflight accepting a split file,
`withAgentSecret` top-up and its warning.

Also made the handshake tests hermetic: they spawned with the runner's real
`$HOME`, so once resolution reads `~/.switch/agents/` a developer's own
credentials would have decided whether they passed.

## Known gap

`.claude/switch-subagents/<name>.settings.json` — the pre-CHOO-1440 legacy
subagent file — is still read with an inline token for un-migrated installs. It
is a fallback path only; nothing writes it any more.

[CHOO-1962]: https://sandboxquantum.atlassian.net/browse/CHOO-1962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHOO-1936]: https://sandboxquantum.atlassian.net/browse/CHOO-1936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amaudruz:
Two changes, then I think this is right.

**1. A startup failure never reaches the user.**

Every failure path here calls `failStartup` — partial env, multi-endpoint, no agents, no store. The process exits before the MCP handshake, and a host reports that as nothing more than a closed connection. The stderr line and `startup-error.log` are technically there, but nobody looks: what the user actually sees is the Switch tools missing, with no reason given.

The runtime is the only thing that knows why it failed, and the agent is the only channel back to the user. So don't exit — start anyway and answer with the error.

Concretely: when resolution fails, serve a single diagnostic tool whose result is the full message you'd have written to stderr — "agents here span two Switch servers: `reviewer` → A, `planner` → B; set `SWITCH_API_ENDPOINT` to the one you meant." The agent reads that and explains it to the user in plain language.

This isn't a retreat from fail-loud, it's louder. `failStartup` is currently the *quietest* possible failure from where the user is standing.

One caveat worth naming: with no credentials there's no catalog to fetch, so this degraded mode can only offer that one tool. Fine — one tool that explains itself beats fifty that silently aren't there.

**2. One place for credentials, not two.**

Let's drop the `$HOME` split and keep the token in `.switch/agents/`, where switchdash already puts it. The working-tree exposure is real, but two credential locations is something everyone then has to hold in their head, and it's currently spread across migration generation 3, `deleteAgent`, the sidecar dual-read and the remote preflight.

Keeping one location should cut this PR by well over half and leave the part the ticket asked for.

**Keep** the connector `.mcp.json` losing its `env` block, though — that's independent of where the token lives. A `${VAR}` resolving to nothing fails the whole MCP server, which breaks the standalone case either way.

I verified the half of that which was load-bearing, since it affects the main path: a stub MCP server declared with **no `env` block at all**, launched by `claude -p` with the `SWITCH_*` vars exported in the parent shell (what switchdash does). All three arrived in the child, along with the full 58-variable environment — so Claude Code passes its environment straight through rather than filtering it, and switchdash's sessions are unaffected by the removal. (claude 2.1.224, via `--mcp-config`; a plugin-bundled `.mcp.json` is a slightly different load path.)

Note this is the opposite of Codex, which hands MCP servers a fixed allowlist — hence `env_vars` naming each variable there.

We can move the secret out of the working tree as its own change later.

christian-mcdermott:
Both changes are in — `dc10ca8` (diagnostic tool) and `2179bcf` (one credential location). Thanks, the first one in particular is a better call than what I had.

**1. Startup failures now reach the session.**

Resolution failure no longer exits. The server completes the handshake and serves exactly one tool, `switch_unavailable`, whose result is the diagnostic. Folded the unreachable-server case in too — from the host's side it is the same closed pipe as having no credentials, so it had the same invisibility.

**Any** tool name answers while degraded, not just its own: a host replaying a cached list would otherwise get `Unknown tool`, which says nothing.

Two things fell out of the change that are worth flagging:

- **The port-file race.** The `${SWITCH_*}` exit existed partly so a pre-expansion spawn could not race the real one for `~/.switch/sessions/<ppid>/port`. Nothing exits now, so degraded mode starts no hook listener and writes no port — there is nothing left to race for. Covered by a test.
- **`unpublishPort()` was deleting the diagnostic.** It clears the whole session directory, and it ran *after* resolution — so the `startup-error.log` written moments earlier was wiped before serving. Invisible while a failure exited on the spot; load-bearing once it survives. Moved ahead of resolution.

**2. One credential location.**

Dropped the `$HOME` split. Gone with it: the home store, `WorkspaceFs.secrets`, the identity/secret reader split, migration generation 3 (marker back to 2, generation-1 skip restored), the sidecar dual-read — and therefore `SIDECAR_VERSION` back to **1.7**, there being no sidecar change left to announce.

**44 files → 21; +2035 → +1281.** Your "well over half" was right.

One judgement call inside it, since it is where this and an earlier instruction meet: **`.claude/settings.local.json` still carries no token.** It is Claude Code's own file, read by every session in the directory, and it only needs to name the agent — a token in both files would itself be two places. So the credential lives in `.switch/agents/<slug>.json` alone. Say if you meant the fuller revert.

`.mcp.json` keeps losing its `env` block, as asked.

**Verified live**, not just in tests — the credential path changed materially, so I re-ran against a real Switch with two really-registered agents: **15/15**, covering the real 47-operation catalog and an authenticated call, `select_agent` binding, and each new degraded path (no store, two servers, unreachable server, no port published). Suites: switchdash **1851**, plugins **89**, core **162**, shared **7**, runtime **39**; typecheck, lint, format clean; `pnpm build` including the sidecar bundle.

**On the deferral.** Moving the token out of the working tree is now unticketed and lives only in this thread. The Jira description still records the split under "Decisions already taken", so as it stands the ticket and the code disagree. Happy to draft the follow-up description for whoever owns ticketing — I'd rather it were tracked than remembered.

**One caveat on your verification**, which you flagged yourself: `--mcp-config` is a different load path from a plugin-bundled `.mcp.json`, and the bundled one is what users get. I have `claude` locally and can close that gap if you'd like it closed before merge.

amaudruz (approved):

PR #142, opened 2026-08-06
feat(discord): map native slash commands to Switch in-room commands (CHOO-1879)

christian-mcdermott:
Discord parity for [CHOO-1100](https://sandboxquantum.atlassian.net/browse/CHOO-1100), which did this for Slack. Jira: [CHOO-1879](https://sandboxquantum.atlassian.net/browse/CHOO-1879).

The Discord bridge had no interaction surface at all — a plain `discord.Client` with only an `on_message` handler and no command tree — so all 20 visible in-room commands were `!`-only there.

Branch is merged up to date with `main`.

## One implementation, two entry points

A slash invocation is translated and handed to the same `self._on_command(InboundCommand(...))` call the typed `!` path already makes, so it lands in the same dispatcher in `bridges/agent/commands.py`. No command is re-implemented, and there is no second mapping table to drift.

## Where Discord forces a divergence

Slack hands the bridge one free-text string that becomes `args` verbatim. Discord registers commands with a **declared signature**, so the argument surface has to be described somewhere. It is described **in the shared registry** — a new `Command.args_spec` — rather than in a Discord-side table that would drift from the handlers. Declared option values are reassembled back into the same positional `@token` string the handlers already parse, so everything downstream is byte-identical to a typed command.

Only 7 of the 20 commands take arguments; the other 13 declare none.

Reassembly rejects two inputs Discord itself considers valid:

- **A value that is not a single mention token** — it would lose everything after the space once tokenised.
- **A value supplied after an earlier optional one was omitted.** Discord submits options by name and allows `role` without `agent` on `run-cmd`, but the positional form cannot express it: `@manager` would slide into the agent slot and answer for the wrong target.

## The ~3-second rule

Met by **ordering, not speed**. A command's result never returns down the interaction call — the admin client answers with a room message that reaches the channel later, *outbound*. So there is nothing to defer *for*, and `defer()` would be worse: it obliges a follow-up within 15 minutes that never comes. The acknowledgement goes out before any Switch work and doubles as the thread root the result is filed under, mirroring the Slack path's visible "Running …" message.

## Guild-scoped registration

- The adapter is **single-guild by construction** — `DiscordConnectionConfig.guild_id` is required and every lookup is scoped to it.
- Guild commands apply **immediately**; global ones propagate for up to an hour.
- Global registration is per-**application**, so on an instance running several Discord bridges it would leak each bridge's commands into the others' guilds, where they could only fail.

Sync is a bulk overwrite on every start, so renames and removals reconcile rather than accumulate.

## Failures are always visible

- A bad argument is refused **ephemerally, before dispatch** — the invoker's typo, so it does not need to be a channel-visible "Running…" that turns into an error.
- A failure **after** the ack rewrites the posted "Running…" message into the error.
- A slash-sync failure at startup is logged and non-fatal (broad catch: a missing scope raises `Forbidden`, a missing application id raises `MissingApplicationID`, which is not an `HTTPException`). Disclosed in the logs, not silent.

## Follow-ups found while reviewing this on a real Discord server

Three defects surfaced during review. All are Discord-bridge bugs; two predate this PR.

**1. Slash descriptions taught the `!` form.** Registry descriptions are written for `!help` and end in a `Usage:` clause naming the `!` invocation, so `/reset` advertised ``Usage: `!reset @agent-name` `` — telling someone who had just typed `/` to type `!`. Only the leading summary is registered now. Nothing is lost: Discord renders the arguments itself, `/reset-all-agents` sits in the same picker, and `!help` still carries the full text. Dropping the clause also takes every description under the 100-char cap on its own.

**2. Messages over 2,000 characters were silently dropped (pre-existing).** Discord rejects them with a 400; both send paths caught it, logged, and returned `None`, so the user saw nothing. Slack, Mattermost and Teams allow 8–20× more, so nothing else in the bridge layer had hit it — but 2,000 characters is very little for an agent reply. `help` is the reproducible case at 2,063 characters, which means `!help` has been failing this way on Discord for as long as the command list has been this long.

  Outbound text is now split to fit, breaking on a line boundary or a word. A break inside a fenced code block closes the fence and reopens it with the same language, since a torn fence renders the whole remainder as plain text. The **first** part's ref is returned and mapped, so replies still thread under the start of a message. A part failing mid-sequence leaves a visible notice rather than a message that just stops.

**3. The no-agents notice advertised Slack's syntax everywhere (pre-existing).** It offered one `/invite-agent @agent-name` line to every platform — wrong on Discord (which names each argument: `agent:agent-name`), and wrong on Mattermost and Teams, which register no slash commands at all. The line is now the adapter's to supply, alongside `_bridge_display_name`; returning `None` drops the bullet rather than pointing at a command that does not exist. Discord reads the option name from the registry the commands are registered from, so renaming it cannot leave the notice stale.

## A bug the tests caught

The first cut used `tree.add_command(cmd)` then `tree.sync(guild=…)`. Without `guild=` those register as *global*, so the guild-scoped sync shipped an **empty payload** — registering nothing at all, silently. The test now pins the synced payload against the registry.

## Setup change reviewers should note

Existing Discord bots need re-inviting with the **`applications.commands`** OAuth2 scope. Without it the bridge runs fine but logs a sync failure and only answers `!`-commands. Documented in `docs/bridges/DISCORD_SETUP.md`.

## Testing

~50 new tests. 647 passing across `core/tests/switch_core/bridges/`, `ruff` and `mypy` clean. Coverage spans registration-vs-registry parity, description rendering, every reassembly edge case, thread/channel resolution, ack ordering, each failure path, chunk splitting/fence handling/round-tripping, and the per-bridge notice.

Store/integration tests needing Docker were not run in this environment.

## Deliberately not folded in

- **A pre-existing bug in the `!` path**: `discord/adapter.py` passes raw content to `_on_command` without `translate_inbound`, so a native `<@123>` mention inside a typed `!command` never resolves. Slack's slash path does translate. The slash path is unaffected — its options are plain strings.

[CHOO-1100]: https://sandboxquantum.atlassian.net/browse/CHOO-1100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHOO-1879]: https://sandboxquantum.atlassian.net/browse/CHOO-1879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amaudruz (approved):

amaudruz (changes requested):
to adapt

Image

amaudruz (approved):

PR #187, opened 2026-08-10
feat: Telegram collaboration bridge (CHOO-1686)

christian-mcdermott:
Adds `telegram` as a collaboration bridge type, at parity with Slack, Mattermost and Discord.

Jira: [CHOO-1686](https://sandboxquantum.atlassian.net/browse/CHOO-1686)

## Shape

`discord/adapter.py` is the template: one `adapter.py` holding both the connection config and the adapter, an empty package `__init__.py`, and a single `register_adapter` call in `main.py`. Section order is copied from Discord.

## The three Telegram-specific calls

Agreed with @cmcdermott before implementation:

- **Single bot.** Telegram has no per-message identity override — nothing like a Discord webhook's username/avatar — so every agent posts through one bot with its name at the head of the message. This is the degradation Discord already falls back to in DMs, applied everywhere.
- **`create_channel` raises.** The Bot API genuinely cannot create a chat. The error names the bot and tells the operator to create the chat and add it, rather than pretending to succeed.
- **`create_dm_channel` keeps the base `NotImplementedError`.** Telegram DMs are initiated by the user; the bot picks the conversation up once they message it.

Transport is long polling, so the bridge needs no public ingress — parity with Slack and Discord rather than Teams.

## Beyond straight parity

Four places where copying Discord would have shipped a worse bridge:

- **`send_attachments` is overridden.** Discord leaves the base loop in place, so a multi-file message becomes N posts. Telegram posts an album via `sendMediaGroup`, falling back to one-per-file for mixed or out-of-range batches.
- **`prime_mention_targets` is implemented.** Discord leaves it as the inherited no-op and only learns a person's id once they have spoken, so mentioning a quiet user degrades to plain text. Telegram has the same mention-by-numeric-id constraint, so the mapping is taken up front.
- **Markdown is converted to Telegram's HTML subset.** MarkdownV2 requires escaping ordinary punctuation and rejects an entire message over one stray character. If Telegram refuses the markup anyway, the message is re-sent unformatted with a warning rather than lost.
- **Oversize payloads are handled, not rejected.** Bodies over the 4096-character cap split on line boundaries; a caption over the 1024-character caption cap posts as its own message. Agent output routinely exceeds both.

Telegram also signals when the bot is added to a chat, which Discord has no equivalent for — so that is wired to the app-join callback and rooms provision on join rather than waiting for a message.

## Registration points outside the bridge directory

- `main.py` — import + one `register_adapter("telegram", …)`. No DB enum, no migration; the gateway schemas are generic and the operator dashboard is fully data-driven off the registry.
- `core/pyproject.toml` — `python-telegram-bot>=21,<22`.
- switchdash — `telegram.svg` (functional, not cosmetic: the sidebar hides a room's "open channel" button when no icon exists), the platform label and setup-doc maps, and `tg:` added to the `openExternal` scheme allowlist.
- `file-logger.ts` — a Telegram bot-token pattern. The token is unprefixed, so like Discord's it must be matched structurally or it reaches the logs whenever it is not behind a `bot_token=` key.
- `.gitleaks.toml` — an allowlist path for the new tests; gitleaks ships a `telegram-bot-api-token` rule that fires on any realistically-shaped fixture.

One drive-by fix: the room titlebar labelled its button from the raw slug, so it read "Open in telegram". It now uses the platform label map like every other call site.

## Docs

`docs/bridges/TELEGRAM_SETUP.md` mirrors `DISCORD_SETUP.md`. It leads on **disabling privacy mode**, which is Telegram's equivalent of Discord's message-content intent — skip it and the bot sees almost no group traffic, which is the kind of failure that costs an afternoon.

Also updated: the bridges README, `ARCHITECTURE.md`, the repo README, `CLAUDE.md`, `.env.example`, and both connector skills (added identically, then diffed to confirm the only remaining differences are the intentional host-specific ones).

No version numbers touched — changelog entries sit under `[Unreleased]`.

## Tests

`test_telegram_adapter.py` (95 tests) and `test_telegram_provisioning.py` (26) to the Discord standard, plus entries in the four shared files that enumerate every bridge — including the `test_home_deeplinks.py` tuple asserting no home link leaks a credential, and the dash test requiring the linked setup guide to exist in the repo.

Writing them turned up a real bug, fixed and covered: when a caption too long for Telegram had been posted as its own message and the upload then failed, the fallback notice posted the same caption a second time.

**Verification:** 1156 backend tests pass against a real PostgreSQL; 2339 switchdash tests pass. `ruff format`/`check` clean, `mypy` adds no new errors (31 pre-existing, 0 in the new code), `tsgo`/oxlint/oxfmt clean, `gitleaks` finds no leaks.

## Note for the reviewer

The ticket's checklist says 14 abstract methods; the base class has **15**. I worked to the code — worth correcting on the ticket.

---

# Review round (Louis's feedback)

Five items from @cmcdermott, mapped from @amaudruz's review. Three of the symptoms had a different cause than they looked like.

## Louis's major point: room creation

Built as the general capability Louis described, not a Telegram patch. Whether a connection will create a channel now has two halves, and both must hold:

- **What the platform can do** — a fixed fact, declared on the adapter class so it is answerable for a stopped bridge and for a type no connection exists for yet. Telegram is the only `False`.
- **What the operator permits** — a per-connection switch set at registration and changeable after, so a deployment can withhold channel creation from Slack or Teams too. It can only narrow the platform's ceiling; registering or patching a connection to claim otherwise is refused.

The answer reaches everyone who acts on it: both room forms stop offering the option and say **which** of the two reasons applies, `list_bridges` carries `can_create_channels`, and Switch Console's empty state tells a Telegram-only deployment what to do instead of showing a bare "no messaging apps".

**The 500 was a side effect worth calling out.** The adapter always had a good explanation. It was being discarded because `NotImplementedError` subclasses `RuntimeError` and every caller of `create_room` catches only `ValueError` — so four of the six doors into room creation returned `Internal Server Error` while the useful text stayed in the server log. It is now a `ValueError` subclass. Opening a DM is governed the same way, and `change_bridge` — which provisions a channel too — is guarded on the same path.

One migration, and it backfills: an existing Telegram connection is switched off on upgrade. Verified against a real Postgres, including the downgrade.

## "Open in Switch Console" doesn't work

It could not have. The link is a `switchdash://` URL and Telegram renders only `http(s)`/`tg:`/`mailto:` — the client drops the link and keeps the label, which is the dead grey text in the screenshot.

Worse: when Telegram rejects a message for that, the plain-text retry only fired if the error contained the word `"parse"`. `"unsupported URL protocol"` does not, so **the entire message was discarded** with a single log line — a working indicator or an `!status` reply could simply vanish. Any rejection is retried unformatted now.

The actual fix is `GATEWAY_PUBLIC_URL`, which rewrites the deeplink to the https redirect and works in the Telegram app and on Telegram Web alike. It is a deployment setting, it was not set, and nothing said so — the bridge now warns at startup, and the docs state it as required rather than a refinement. Unset, the address is posted as tap-to-copy text rather than as a link the platform silently eats.

## `/invite_agent`, and the notice around it

Louis is right, and the reason is that Telegram will not register a command containing a hyphen — so `/invite-agent` is the one spelling its menu never autocompletes. All three forms are listed now, underscore first.

**The asterisks in his screenshot were a second bug.** That notice is written in Switch Markdown like every other body, but `admin_message` handed it to Telegram without the conversion every other body gets, under `parse_mode=HTML`. Fixed by converting there, and by rewriting the adapter's own hand-rolled HTML notices as Markdown so there is one content format at the boundary instead of two. Two notices also used single-asterisk emphasis, which is not Switch's Markdown — Slack bolded them by accident, Telegram italicised them.

## The icon

A real bug, not taste: the plane was a hole punched through the disc rather than a white shape, so it looked right on white and came out **black** on Console's dark theme. Now an explicit white plane, identical on both.

## Telling agents apart

Each agent's messages carry a stable coloured mark beside the name, derived from the name by hash — no state, no configuration, same colour after a restart.

**Deliberately not provider logos.** Switch only distinguishes `claude-code` from `codex`, and Switch Console registers every other provider — Gemini, Cursor, the rest — as `claude-code` (its own code comments say so). A logo drawn from that would confidently mislabel most agents. Worth doing once that data can support it.

## Docs

Louis's second comment: creating a chat and adding the bot now **leads**, as step 2, with the dashboard's one-click link after it as the shortcut it is. Channel creation is documented in the shared bridges page, since it applies to every platform.

## Verification

1345 backend tests against real Postgres · 2413 Switch Console tests · gateway typecheck + build · `ruff`/`mypy` clean (31 pre-existing mypy errors, none added) · oxlint clean (1 pre-existing warning in an untouched file) · `gitleaks` no leaks · migration applied, backfilled and reversed against real Postgres.

**Still outstanding, and not mine to do:** both connector plugins have changed on this branch and the releaser has since bumped them on main, so these skill changes sit unreleased on top. They need a bump before merge or installs will not pick them up.

[CHOO-1686]: https://sandboxquantum.atlassian.net/browse/CHOO-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amaudruz:
Nice ! Minor feedback:

image

the icon looks pretty bad in Switch Console, not sure how much we can actually make it better


image

it should be "/invite_agent" instead of "/invite-agent"

image

It can be hard to identify which agent is sending which message. I understand that we can't overwrite the app name and icon, but maybe we could change the formatting a bit to make it more obvious  ? bonus point -> could we have logos for the agent provide ?  

Major feedback:
Given that we can't programmatically create rooms in Telegram, this should be reflected in multiple places, examples:
- when agent agent tries to create a room it fails and it does not know why
- when create a room from the gateway or switch console it also fails with cryptic error

Both the agent the uis need to know that we can't do this for telegram

That could actually be a general cross messaging app concept, where when we onboarid a messaging app (through gateway or console) we need to specify if the connection can or not create rooms. In some cases companies might not waht the connection from Slack or Teams to allow for creation rooms, and the switch bot might not get the permissions for it. And for telegram you would simply not allow to say yes.


image
The "Open in Switch Console" link does not work

amaudruz:
one more comment, in the telegram docs it should provide as first approach to have a room linked to telegram by simply creating a chat and inviting the bot in it

amaudruz (approved):

PR #190, opened 2026-08-10
refactor: rename switchdash identifiers to switch-console, and dash/ to console/ — CHOO-2008

amaudruz:
Re-opens #181 against `main`. #181 was merged into this branch rather than `main`, because it was created as a stacked PR while this branch was still its base — so its content never reached `main`.

Nothing has been re-done: this is the same reviewed content, plus a merge commit recording that `main` is accounted for. The tree is unchanged by that commit.

## What is in here

The identifier half of the rename — everything #178 (display name) deliberately left alone:

- workspace packages → `@switch-console/*`, app directory → `apps/switch-console-desktop/`
- release artifacts, tags (`switch-console-v*`) and the Windows signing profile
- the artifact-registry key, and every name that lives only inside this repo
- two latent bugs found on the way: the release workflow filtered on a package name that never existed, and `resolveAppVersion` matched a `package.json` name that never existed either
- a rename-induced break: `safeStorage` keys its encryption on the name the app announces to the OS, so a renamed build could not read existing secrets. `OS_APP_NAME` is now frozen on `Switchdash` (measured on macOS, see `console/AGENTS.md`), and an unreadable session cookie reports as signed-out rather than "cannot reach the server"

Deliberately still `switchdash`: the user-data directory and database, the app id, the deeplink scheme, `.switchdash.json` / `.switchdash/`, and the `SWITCHDASH_*` env vars and `X-Switchdash-*` headers — all of them are already on users' disks or in deployed sidecars. The reasoning is written down in `console/AGENTS.md`.

## Verification

Typecheck clean; 2305 tests pass; artifact registry check agrees.

**Not verified here:** that a packaged macOS build reads its pre-rename secrets. That needs a real launch on a Mac, which a green suite cannot substitute for.


amaudruz:
Added on top: the root folder `dash/` → `console/` (one pure-rename commit, then a small commit pointing CI, the release workflows, the artifact registry and the root docs at the new path).

Folded in here rather than opened as a fourth stacked PR — stacking is what put #181 on the wrong base. Splitting it back out is one command if you'd rather review it separately.

Measured the cost to in-flight branches rather than guessing: merging this into #187 (telegram) gives **1** advisory conflict where git has already moved the added file correctly, and into #182 (feedback removal) **0**. Edits follow the rename automatically. `merge.directoryRenames=true` removes even the advisory.

`CHANGELOG.md` deliberately untouched — its entries record where things were when released.

PR #201, opened 2026-08-11
Scrub internal references now the repo is public (CHOO-2023)

amaudruz:
Follow-up to #197. That PR removed the private-repo *machinery*; this is the cosmetic half — what a stranger reads when they land on the repo. No behaviour changes; docs, the changelog and the justfile only.

### Ticket keys
Dropped from `CHANGELOG.md` (159 of them) and `console/docs/INSTALL.md`, where they point at a tracker readers cannot open. **Kept** in the ~275 source comments and the design notes, where they are real traceability for whoever maintains this.

The root `CLAUDE.md` now writes that split down, in a new "This repository is public" section covering credentials, internal hostnames and infrastructure, personal addresses, and internal tooling — with a pointer to it from `console/AGENTS.md`.

The internal audit finding ids (`H5`, `M5`, `M1`) go for the same reason: nobody outside can resolve them. The PR numbers stay, since those do resolve.

### justfile
The optional `import? 'internal/justfile'` is removed. It degraded gracefully, but `internal/` is untracked and never ships, so all the line did was advertise AWS/EKS tooling no reader has.

### Front door
- `CODE_OF_CONDUCT.md` — Contributor Covenant 2.1. **The enforcement contact is a personal address; swap it for a role alias when one exists.**
- Linked it, and `SECURITY.md`, from `README.md` and `CONTRIBUTING.md`.
- The two "coming soon" badges linked to `#`. Docs now points at `docs/`; the website badge is gone until there is a website.

### Left alone deliberately
- **Connector skills** — the `CHOO-509` there is an illustrative worklog example, not a stale pointer. Editing them would also oblige a plugin version bump, which is the releaser's call.
- **`docs/api/AGENT_PROTOCOL.md`** — the ticket keys are the subject of the text (it has a section on which tickets the design closes). Stripping them would damage the document.
- **`.github/renovate-security/*`** and the `maintainer:` email in the electron-builder configs — flagged in CHOO-2023, confirmed out of scope for now.

### Review note
The changelog diff is one line per removal with no re-wrapping, so some paragraphs are now raggedly wrapped. That keeps the diff verifiable — every changed line should differ only by a deleted ticket key. Markdown renders identically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project aims to adhere to [Semantic Versioning](https://semver.org/spec
 
 Switch ships several independently-versioned artifacts, each with its own
 section below. Every artifact carries three-part semver (`MAJOR.MINOR.PATCH`)
-and a changelog, without exception (CHOO-1865).
+and a changelog, without exception.
 
 A version says **where an artifact is** — which release you are running. It says
 nothing about what that release can talk to. Compatibility is carried separately
@@ -46,7 +46,7 @@ version of their own to them without also giving them a release of their own.
 #### Added
 - The Discord collaboration bridge registers native slash commands (`/…`) that
   map onto the same in-room command dispatcher as typed `!` commands, giving
-  Discord the interaction surface Slack already had (CHOO-1879). One
+  Discord the interaction surface Slack already had. One
   implementation, two entry points: a slash invocation is reassembled into the
   same positional `@token` string the `!` handlers already parse. Argument shapes
   come from a new shared `Command.args_spec` in the command registry — validated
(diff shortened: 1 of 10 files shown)


amaudruz:
### Added: `docs/` staleness pass

Second commit audits `docs/ARCHITECTURE.md` and `docs/api/AGENT_PROTOCOL.md` claim-by-claim against the code (`docs/bridges/` excluded). Every correction below was verified against the source, not inferred.

Neither file had suffered from the `console/` rename or the private-repo scrub — no dead paths, no orphaned GitHub-CLI references. The rot was elsewhere.

**ARCHITECTURE.md**
- `PUT /agents/{id}/addressing-policy` does not exist. It's `PUT /gateway/agents/{id}/addressing-policy`, cookie-JWT and owner-or-admin — following the doc got you a 401.
- `/gateway/collaborations` is admin-gated for **writes only**; reads take any authenticated user.
- `/mcp` accepts a **third** credential kind, an OIDC token, when `OAUTH_ISSUER_URL` is set. This is what the otherwise-unexplained `/oauth` bypass prefix is for.
- `handle_protection_verdict` was described as "gated pending protection setup". It has no callers at all, and it's in the *collaboration* bridge, not the agent bridge.
- Missing entirely: the server-side connector subsystem (`ServerConnector`, `/gateway/connectors`, the OpenCode connector), the `/agents/{id}/ops` operation surface, `/version`, `/deeplink/session`, and `addressing.py` — which the doc itself tells you to go read.

**AGENT_PROTOCOL.md**
- The header still said "design, under review" for a design whose Stages A and B have shipped end to end. That one line makes a reader discount the whole document.
- §9.1 was the worst-affected section: the runtime is published to npmjs, not copied into each plugin subtree; both hosts register it from a bundled `.mcp.json`; the Codex profile carries specialization only and registers no server (the doc contradicted itself on this).
- A failed operation fetch **degrades** to a single `switch_unavailable` tool — it no longer refuses to start. The tool surface is also the server registry *plus* three runtime-local tools.
- `SWITCH_CHANNEL_DISABLE_POLL` was predicted to be deletable. It was repurposed instead, keeping the name — now suppresses notification surfacing, not polling.
- Stage B item 5 had shipped but carried no marker; three of §1's four problems are fixed and now say so.
- Corrected in both directions: CHOO-1810 was claimed closed and **is not** (both notification builders still exist), and `room_name` is marked proposed rather than "new", since it isn't implemented.
- Payload field lists for `connection_state` and `gap` were incomplete.

Version numbers in illustrative payloads are left alone deliberately, as is everything else version-shaped.

PR #203, opened 2026-08-11
feat: support opencode agents in Switch Console (CHOO-1103)

christian-mcdermott:
Makes OpenCode a first-class Switch agent type, locally and on remote hosts.
Also carries **CHOO-2057** and a fix for the OpenCode permission hang, both
folded in at the requester's direction.

**Complete and ready for review.** #186 has landed, so the third Switch skill
is in and this PR no longer depends on anything.

A verification plan for this PR is in the first comment below.

## Where this came from

This ticket was worked once before, on `feature-request/opencode-switchdash-support`
in the pre-cutover repository, and never carried over. That branch was read
before anything was written here. The triage call was **adapt with reference**
rather than port or re-derive:

- Ported close to as-is: the permission-hang fix, the session-control recipes,
  the Linux user-prefix install.
- Adapted: the turn cycle and tool activity, landed as their final state rather
  than the sequence of intermediate forms, all of which are known-broken.
- Re-derived: the connector and everything around credentials. The old branch
  wrote the bearer token into a config file, which is exactly what the
  provider-neutral credential work has since removed. Porting its
  credential-resolution-by-agent-type would have been a regression, not a fix —
  that mechanism is already provider-neutral, so the work there was verifying
  OpenCode goes through it, not adding a dispatch.

Two behaviours from that branch would have cost real time to rediscover, and
both are now pinned by tests: OpenCode re-emits the user message ~20ms *after*
`session.idle`, so the obvious turn-start check bounces the session back to
working permanently; and its interrupt needs Escape twice as two separate
writes, because two ESC bytes in one write are coalesced and silently ignored.

## The one design decision

Both existing connectors install by driving the host's plugin-marketplace CLI.
**OpenCode has no such CLI** — `opencode plugin <module>` installs a single npm
module and has no list, remove or version verb — so neither dialect applies and
`switchSetup: 'none'` meant no Switch tools at all.

`switchSetup` gains a `files` kind: the plugin declares that its connector is
written rather than fetched and supplies the behavior that writes it, reaching
the same status/install/update/uninstall surface through a home-rooted
`PluginFs` so one implementation serves local and SSH alike.

For OpenCode the connector registers the Switch MCP server in its global config
as a **local (stdio)** server. That is what keeps the credential off disk:
OpenCode spawns local servers with the parent environment, so the runtime
inherits `SWITCH_*` from the session and nothing secret is written into a file
every OpenCode session on the machine reads. This was verified against the
shipped binary, not assumed — including that OpenCode *does* interpolate
`{env:}` and `{file:}` (so a header-carrying remote server was also possible;
stdio was chosen because it needs no interpolation), and that its config rejects
unknown keys on an MCP entry and fails the whole config with them, which is why
the install bookkeeping sits beside the config rather than inside it.

`connectors/opencode-plugin/` is the source of truth for the connector's
assets, matching the other two: manifest, MCP fragment, reporting plugin,
version, `artifacts.yaml` entry, changelog section. The remaining difference is
that nothing *installs* it — Switch Console writes the files — so it is not in
the marketplace manifest, and an install reports the app version that wrote it.
Publishing to npm would close that gap; it is a separate decision, since it
means a new released artifact.

Because the app writes the files it also embeds them, and the two copies can
drift with nothing to notice. `connector-assets.test.ts` compares them byte for
byte — the directory is what gets reviewed while sessions run what is embedded,
so shipping stale behaviour would look, in review, exactly like shipping the fix.

## Also in here

- **CHOO-2057** — the Add Agent picker auto-selected only when exactly one agent
  type was usable. Harmless until the whole lower half of the dialog was gated on
  a chosen type, at which point a second connector turned onboarding into a form
  that renders one line of grey text. It now selects the only usable type, or the
  user's configured default when that is usable on the target machine, and is
  blind to how many types exist so a third does not move it again.
- **OpenCode permission hang** — the server-side connector ignored
  `permission.updated`, so a tool permission request was never answered, the
  session never went idle, and the response stream blocked forever.

Three further defects surfaced while working and are fixed here, since OpenCode
hits all three: neither remote path would install hooks delivered as a dropped
plugin, the Linux install needed root, and resume fell through to `--continue`,
which reattaches to whatever ran last in that directory — for a Switch agent,
possibly another agent's conversation.

## What is not in this PR

- **No version bumps.** Versioning is the releaser's call on this branch. Note
  the Claude Code and Codex skills *were* edited — `opencode` is now a value
  `list_agents` filters on and a key `update_agent_detail` takes options for, so
  all three skills say so — which normally requires bumping both plugin
  manifests. **Both need a bump before release or the change will not reach
  installs.** The new connector starts at `0.1.0`.
- Several other providers declare `sessionIdOnResumeOnly` and are absent from the
  resume list this PR adds OpenCode to. Whether they have the same problem is not
  established, so they are left alone rather than changed untested.

## Testing

`format`, `lint`, `typecheck`, `just artifacts-check` clean. 2,279 console tests
pass (`node` + `main-db`), 139 plugin-package tests, 970 Python tests. mypy
reports 31 errors, all pre-existing on `main` and none in touched files.

**Two gaps, both real.** Python store tests could not run locally — Docker was
unavailable, so 147 testcontainer-backed tests error at setup and need
confirming green in CI. And none of this has been exercised against a real
OpenCode session yet; that is what the verification plan is for.

Merged `main` (including #186) into the branch; the only conflict was
`CHANGELOG.md`, resolved by keeping both sides.


christian-mcdermott:
# CHOO-1103 — verification plan

Covers PR #203: OpenCode agent support, plus CHOO-2057 and the OpenCode
permission hang.

Nothing here has been run against a real OpenCode session yet — that is the
whole point of this plan. Sections **A–D** are the ones that decide readiness.
**E** is the regression surface: this PR changes shared code paths, so Claude
Code and Codex have to be re-checked even though nothing about them was the
target.

Record each numbered check as pass / fail / not-run. A "not-run" is a known gap,
not a pass.

---

## 0. Preconditions

| | |
|---|---|
| Branch | `feature-request/opencode-support` (PR #203) |
| OpenCode | `npm install -g opencode-ai`, verified with `opencode --version`. Pinned behaviour below was read from **1.18.16** — note the version actually tested |
| Switch | a reachable instance, and a room the test agent can be added to |
| Console | built from the branch (`pnpm install && pnpm run dev` from `console/`) |
| Remote (section D only) | an SSH host reachable from Switch Console, with Node and no root for the agent user — the no-root part is the point |

**Two observation points used throughout:**

- **OpenCode's own log** — `~/.local/share/opencode/log/`. The connector plugin
  logs here and nowhere else. A line `plugin loaded {build: turn-cycle-v4-tool-activity}`
  confirms which build a session actually has, which matters because the plugin
  file is copied into working directories and can go stale.
- **The room** — for anything user-visible, the room is the source of truth, not
  the console UI. A session can look right locally and be reporting nothing.

**Before starting**, capture a baseline: `~/.config/opencode/opencode.json` (if
it exists) and any `.opencode/` in the test working directory. Several checks
assert these are left alone or restored.

---

## A. Connector install — the part with no precedent

The other two connectors are installed by their host's marketplace CLI. This one
is written by Switch Console, so this section is where a novel mistake is most
likely to be.

**A1 — OpenCode appears as a Switch agent type.** Settings → Agents lists
OpenCode with a Switch connector card, alongside Claude Code and Codex.
*Previously it appeared as a CLI provider only, with no Switch surface at all.*

**A2 — Install writes the MCP server.** Click Install. Then
`~/.config/opencode/opencode.json` contains an `mcp.switch` entry of
`type: "local"`, `command: ["npx","-y","@sandboxaq/switch-agent-runtime@0.3.0"]`,
`enabled: true`, `timeout: 60000` — and **nothing else** in that entry.
> Unknown keys are not cosmetic: OpenCode rejects them and fails the *whole*
> config, so the agent would not start.

**A3 — No credential on disk.** `grep -riE 'token|bearer|authorization' ~/.config/opencode/opencode.json`
returns nothing. This is the central security claim of the design.

**A4 — Existing config survives.** Put an unrelated MCP server and an unrelated
top-level setting (e.g. `"theme"`) in `opencode.json` first. After install both
are still there, unmodified.

**A5 — Status is honest.** The card shows installed. Now delete the `mcp.switch`
entry by hand and re-read status: it must report **not installed**, not
installed-and-fine.
> The install marker is a separate file; trusting it alone would report a
> connector that is not actually registered, hiding exactly why the agent has no
> Switch tools.

**A6 — Idempotent.** Install twice. The config is unchanged the second time.

**A7 — Uninstall is surgical.** Uninstall. `mcp.switch` and the marker file are
gone; the unrelated server and setting from A4 remain.

**A8 — A broken config fails loudly.** Put invalid JSON in `opencode.json` and
install. It must refuse with a readable error and **leave the file untouched** —
not overwrite it.

---

## B. A session in a room

Reinstall the connector, onboard an OpenCode agent against a working directory,
and add it to a room.

**B1 — Onboarding registers the right type.** After onboarding, the agent's
detail in Switch shows connector type **OpenCode CLI**, not Claude Code.
*This was wrong before: OpenCode agents registered as `claude-code`.*

**B2 — The agent starts and connects.** Start a session. It joins the room and
appears as a participant.

**B3 — The Switch tools are actually there.** In the session, confirm the
`switch` MCP server is connected and its tools are callable — ask the agent to
list its rooms. This is the end-to-end proof that A2 produced a working
registration and that the runtime got its credentials from the environment.
> If this fails, check the session's environment for `SWITCH_API_TOKEN` before
> anything else — the whole stdio design rests on inheritance.

**B4 — The plugin loaded.** OpenCode's log has the `plugin loaded` line with the
build tag.

**B5 — Addressed message → reply in the room.** Address the agent from the room
(or its bridged channel). It responds *in the room*.

**B6 — Status goes working, then completed.** While the turn runs the agent
shows working; when it finishes, completed.
> *The regression this replaced:* it used to sit on "awaiting input" for its
> whole life. Watch specifically for it flipping **back** to working a moment
> after finishing — OpenCode re-emits the user message ~20ms after going idle,
> and that is the bug the message-id gate exists to prevent. Give it ten seconds
> after the reply before calling this a pass.

**B7 — Tool activity names the step.** During a turn that edits a file and runs
a command, the bridged "working on it" message updates in place with lines like
`Running tool `edit` — foo.py` and `Running tool `bash` — <command>`.

**B8 — Sub-agent spawns are not reported.** If the agent uses its `task` tool,
that does **not** appear as an activity line.

**B9 — Session id captured.** The session's provider session id is recorded and
starts with `ses`.

---

## C. Room control and resume

**C1 — Interrupt.** Start a long turn, then `!interrupt` from the room. The turn
stops.
> The single most likely silent failure in this PR. OpenCode needs Escape
> **twice**, as two separate writes; sent together they are coalesced and
> ignored with no error anywhere. If the turn keeps running, this is that bug.

**C2 — Compact.** `!compact` — the context is compacted and the agent posts back
in the room that it is done, having reconnected and re-assumed any role.

**C3 — Reset.** `!reset` — a fresh session (`/new`, not `/clear`, which OpenCode
does not have), then the same reconnect-and-announce.

**C4 — Room reports the capabilities.** Before any of the above, the room shows
reset / compact / interrupt as supported for this agent. *They were reported
unsupported before.*

**C5 — Resume attaches to the right conversation.** Stop and resume a session
with a captured provider session id: it resumes **that** conversation.

**C6 — Resume with no captured id starts fresh.** Clear the stored provider
session id and resume. It must start a **new** session.
> Not a cosmetic distinction: the fallback used to be OpenCode's `--continue`,
> which reattaches to whatever ran last in that directory — potentially another
> agent's conversation entirely.

---

## D. Remote

Everything in D was broken before this PR; there is no "still works" baseline.

**D1 — Host setup offers OpenCode.** The remote host's setup plan includes
OpenCode as an installable agent type.

**D2 — Install needs no root.** Install OpenCode on the host as a non-root user.
It installs under `$HOME/.local` and does not fail with EACCES, and `opencode`
is on the PATH afterwards.

**D3 — Remote connector install.** Install the Switch connector for OpenCode on
that host; `~/.config/opencode/opencode.json` on the **host** gets the same entry
as A2, with no credential.

**D4 — Remote session reports.** Start a remote OpenCode session in a room.
Repeat **B5–B7** against it.
> This is the check that covers the sidecar/SSH hook install. Before this PR
> neither remote path installed a plugin-delivered hook at all, so the symptom
> of a regression here is a session that answers normally but shows working
> forever.

**D5 — Auto-session.** Configure the agent with auto-session, ensure no live
session, and address it from the room. A session is spawned on the host, joins,
and answers.

**D6 — Remote status and uninstall.** Status reads correctly for the host, and
uninstall removes the entry there.

---

## E. Regression surface

This PR changes code shared by all agent types. None of the following was the
target, and all of it is reachable from what changed.

**E1 — CHOO-2057, the actual fix.** With **two or more** connectors installed,
open Add Agent. The form must be usable: pick a directory and the scan runs, the
onboard-existing list renders, submit is reachable.
> Before: the picker auto-selected only when exactly one type was available, so a
> second connector left it unset and everything below it inert.

**E2 — The auto-select rule.** Walk the matrix:
- exactly one connector installed → that type is selected automatically;
- two or more, and the configured default agent is among them → the default is selected;
- two or more, and the default is *not* among them → nothing is selected, and the user can pick;
- none installed → the "no agent type is set up" notice, naming the host if remote.

**E3 — Three types do not change it.** Repeat E2 with all three connectors
installed. Behaviour is identical to two. *The old rule broke on the 1→2
transition; the new one must not have an equivalent 2→3 cliff.*

**E4 — Claude Code unaffected.** Install/status/update/uninstall its connector;
start a session in a room; check status transitions and tool activity;
interrupt, compact, reset.

**E5 — Codex unaffected.** The same, including its per-agent profile and that
its slash commands are still preceded by an interrupt.

**E6 — Credential teardown.** Delete an agent of each type and confirm no
`SWITCH_*` credentials are left behind.
> The behavior interface that owns this changed shape in this PR.

**E7 — Permission hang.** Against the server-side OpenCode connector (the Python
one, unrelated to Switch Console), drive a tool call that raises a permission
request. The session completes instead of hanging, and the log shows the
auto-approval.

---

## F. Automated gates

Run from the branch. All of these currently pass locally except where noted.

```bash
# Python
uv run --project core ruff format --check core/ && uv run --project core ruff check core/
uv run --project core mypy core/switch_core/          # 31 errors, all pre-existing on main
just test                                              # needs Docker — see below

# Console
cd console && pnpm install
pnpm -r --filter "./packages/**" run build             # required before typecheck resolves
pnpm run format:check && pnpm run lint && pnpm run typecheck
cd apps/switch-console-desktop && pnpm vitest run --project node --project main-db
cd ../../packages/plugins && pnpm vitest run

# Artifact registry
just artifacts-check
```

**The one real gap: Python store tests need Docker**, which was unavailable on
the machine this was built on. 970 tests pass; 147 testcontainer-backed tests
error at setup and have **not** been run. They should be green in CI — confirm
that before sign-off rather than assuming, since this PR touches
`core/switch_core/gateway/known_agents.py`.

Skip the `browser` vitest project (needs a real Chromium; CI is its gate).

---

## G. Sign-off

Ready when A–D pass, E shows no regression, and F is green **including the
Docker-dependent tests in CI**.

Two things are deliberately outstanding and are not covered by any check above:

1. **The third Switch skill is not in this PR** — held for #186, which rewrites
   the other two wholesale. Until it lands, an OpenCode agent has the Switch
   *tools* but not the room-workflow *instructions*, so expect it to use the
   tools less fluently than a Claude Code or Codex agent. That is the missing
   commit, not a bug to chase.
2. **Nothing installs the connector but Switch Console.** If install parity with
   the other two matters, the connector needs publishing to npm so
   `opencode plugin …` can install it — a separate decision, since it means a
   new released artifact.

amaudruz:
Nice work ! minor feedback, to be addressed in this PR:

feedback:
image the icon is too big compared to the others

image
I installed the switch opencode  connector but it provides the version of the claude code connector 

image When changing the theme of the switch console UI it does not adapt the opencode session

In the sessions in switch console of opencode I can't scroll down or up 

Major feedback that can be addressed in a subsequent PR:

We need to support specialising opencode agents with instructions, what model to use, effort level, tool restirctions, etc. Just like we do for claude code and codex with advacned settings

amaudruz (approved):

PR #218, opened 2026-08-13
feat(opencode): give OpenCode agents an advanced configuration (CHOO-2125)

christian-mcdermott:
Gives an OpenCode agent the per-agent configuration Codex got in CHOO-1985 — model, reasoning variant, temperature, top-p, step limit, web search and instructions, set at creation and editable afterwards — and stops the machinery underneath it being Codex-shaped.

Jira: [CHOO-2125](https://sandboxquantum.atlassian.net/browse/CHOO-2125). Builds on #203.

## What OpenCode actually does

Established by running opencode 1.18.18 and reading back its own resolved config and agent, not by analogy with Codex. It differs in ways that decide the design:

- **`OPENCODE_CONFIG`, not a flag.** There is no `--profile` or `--config` on argv; that variable is the only way to add a config to a session.
- **The config layers.** OpenCode deep-merges what the variable names onto the config it already loaded, so the Switch MCP server the connector writes survives untouched — and nothing here restates it, which would put a second copy of the runtime pin somewhere it could drift.
- **Reasoning effort is a per-model `variant`.** The accepted values come from the chosen model's capabilities, so it is guided free text rather than a menu. A dropdown needs the form to ask the installed binary what the chosen model supports, which the fixed field lists cannot do — worth doing, not here.
- **No verbosity, no reasoning summary.** They do not exist as OpenCode settings, so they are not offered.
- **Web search is a tool permission**, and is written as one.
- **Temperature, top-p and a step cap** exist here and had nowhere to live in the Codex-shaped config.
- **Instructions are files.** The per-agent `prompt` takes text but *supplies* the agent's system prompt rather than adding to it, which for a coding agent means losing the operating instructions it needs. Top-level `instructions` is the additive path, so the body is a second file referenced by absolute path.

Two silent failure modes shaped the writer, both observed: OpenCode ignores keys it does not recognise without complaining, and an unrecognised key inside a section drops that whole section — including, for `mcp`, the Switch tools. So the writer is driven off OpenCode's declared settings rather than the collected keys.

## The shape change

`AgentProviderConfig` was six columns named after Codex. OpenCode fits none of them cleanly and adds three of its own, so it becomes `{providerId, values}` keyed by the field keys each provider declares. It is a versioned JSON column, so this is **a schema version and an upgrade function, not a migration** — a stored v1 row folds into `{providerId: 'codex', …}` on read, and existing Codex agents keep their settings.

`SwitchLaunchProfile` also gains an `env`, since OpenCode has no argv to load a config with. Env must name the file absolutely while a profile's files are home-relative, and the builder cannot resolve that — for a remote auto-session it runs on the desktop, which does not know the VM's home. So a profile writes a home placeholder and each launch surface substitutes its own; a substitution missed on the remote path fails the spawn rather than handing the agent a literal path.

Also un-hardcodes the two places the UI asked "is this Codex?" — the create-agent form and the Settings tab's restart affordance — both of which were `providerId === 'codex'` and would have silently excluded OpenCode.

## Verification

Beyond the unit tests, the emitted profile was materialized to disk and read back through the real binary: model, variant, temperature, top-p, step limit and the web-search permission all resolve onto `build`, the instructions file is picked up, and the Switch MCP server is still registered.

**One thing I could not confirm:** that the instructions text reaches the model's system prompt. OpenCode resolves the file and `instructions` is the documented additive mechanism, but I did not get a live model request intercepted to see it land, and I would rather flag that than claim it.

Full suite green: 2,435 desktop tests, 179 plugin tests.

[CHOO-2125]: https://sandboxquantum.atlassian.net/browse/CHOO-2125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amaudruz:
image

When selecting model it then replaces the selection by a dict, not good.

Also the model selection does not work on my end, have you tested it ? Maybe due to the yellow warnings, which appear on every model I select...

we don't have to necessarily provide a list of models available if that is too complicated to setup, we can just expect the user to put the right model in.

christian-mcdermott:
Good catch — both symptoms were one bug, and it's fixed in `d891b530`.

**What was wrong.** The combobox fills its input by stringifying whichever item it was handed, deriving that string from the item's `label`, else its `value`, else by serializing it. I was passing `{id, variants}` objects, which have neither — so picking a model wrote `{"id":"ollama/gemma4:latest","variants":[]}` into the field. That string then never matched anything in the catalogue, which is why *every* model you picked went yellow. The wrapper deliberately omits `itemToStringLabel`, so there's no hook to correct it with; the item has to be the string the field stores. The list is now model-id strings, with variants looked up for display.

**The warnings.** Also softened separately: the field is checked on every keystroke, so it was warning while you were still typing a perfectly good name. Anything that's a prefix of a real model is now left alone, and the warning waits until the value can't become one.

**On my testing — you were right to ask.** I did test it, but not properly. The browser test I wrote asserted the field was a text input, displayed its value, and reported typed text. All three stayed true while selection was completely broken, because the test never opened the list. The new test drives the pointer sequence the combobox actually opens on, asserts `aria-expanded`, and clicks a real option. I ran it against the old component first and it fails with exactly the JSON dict from your screenshot, so it's pinning the bug rather than the fix.

**On dropping the list:** I'd rather keep it now it works — it's the same call we already make to populate the reasoning variants, so it costs nothing extra, and the variant dropdown depends on that data regardless. But if you'd still prefer a plain text box, say so and I'll pull the picker out; it's a small change.

amaudruz (approved):

PR #222, opened 2026-08-14
fix(console): stop CLI updates reading as a fault, and unstick onboarding

amaudruz:
Reported from a first-run: a connector installed fine, the CLI had a pending update the user did not want, and the next onboarding steps were greyed out. The update looked like the gate.

## The gate was a stale cache, not the update

Nothing anywhere treats a pending CLI update as "not set up" — the onboarding predicate, the agent-type picker and the readiness rules all key off connector-installed.

What actually happens: the roster of onboardable agent types is cached under `['switch-setup', 'agent-type-availability', …]`, and the connector mutations invalidated `['switch-setup', <agentId>]`, which does not match it. The checklist reads the roster to decide the agent-providers step, and locks every later step behind the first unfinished one — so a successful install left steps 3 and 4 disabled until a window focus happened to refetch. Intermittent, and with nothing on screen naming the cause.

Both mutations now invalidate the roster too. `switch-setup-invalidation.test.tsx` covers it and fails without the fix (verified by reverting).

## The CLI update no longer reads as a fault

It was amber, and it sat *ahead* of the green "Installed" pill — so the nearest explanation for a stuck checklist was an agent that appeared to need attention. Per the ask, a newer agent CLI is now:

- **gone** from the settings agents list, the remote-host agent-type summary row, and the sidebar trouble indicator — all three answer "does this want attention", and it does not;
- **kept** on the agent's own page, demoted from an amber "Update available" card to a plain "Newer version available" with the command that installs it.

Per-step rows inside a host's setup detail still show it, with the button that performs it. That is the surface you reach by going to look.

## The connector keeps its badge, and gains a name

`UpdateAvailableBadge` → `ConnectorUpdateBadge`, reading **"Connector update"**. One badge used to cover both updates and could not say which was behind; the connector is ours and worth acting on.

Note "Update available" still means two further things elsewhere — the sidecar, and Switch Console itself. Out of scope here, but they should not keep sharing the label.

### Checks

Typecheck clean, lint clean bar one pre-existing warning, 268 files / 2451 tests passing.

No version numbers touched; changelog entry under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -387,6 +387,23 @@ version of their own to them without also giving them a release of their own.
 
 ### [Unreleased]
 
+#### Fixed
+- Installing a Switch connector now finishes the onboarding step it belongs to.
+  The list of agent types you can onboard is cached separately from the agent's
+  own status, and the install refreshed everything but that — so the step stayed
+  unticked, and because the checklist locks each step behind the one above it,
+  the rest of onboarding stayed greyed out with nothing explaining why.
+
+#### Changed
+- An update to an agent's own CLI is no longer reported as though something were
+  wrong. It never gated anything, but an amber badge ahead of "Installed" read
+  like a fault on an agent that worked, and it was the nearest explanation for
+  onboarding appearing stuck. A newer CLI is now mentioned only on the agent's
+  own page, in plain text with the command that installs it.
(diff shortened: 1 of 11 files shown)


amaudruz:
Folded into #220 as requested — same commit, cherry-picked onto that branch.

PR #227, opened 2026-08-14
feat(opencode): standalone install — a published connector and a configure skill (CHOO-2127)

christian-mcdermott:
An OpenCode session can now reach Switch with no Switch Console involved.

Two things stood in the way, and the second was not obvious.

**There was nothing to install.** OpenCode has no plugin marketplace, so unlike
the Claude and Codex connectors — which their hosts fetch straight out of this
repository — this one was only ever written onto disk by Switch Console, out of
text embedded in the app. OpenCode's plugin resolver takes exactly two things, a
directory on disk or an npm package name, which leaves a registry as the only
way to install it without cloning this repository. So the package is scoped to
match the runtime, is no longer private, and has a release workflow keyed to its
own tag.

**Installing the package is not installing the connector.** npm leaves the
module in a cache, and OpenCode discovers a skill only as a file in one of a few
directories it searches — a cache is not one of them. So the connector carries
its own install command. It merges the MCP entry into the user's global config,
writes the skills where OpenCode looks for them, records what it wrote, and
reverses cleanly. Everything comes from the files in the connector directory, so
there is no second copy to keep in step.

Then the ticket's actual subject: a **`configure` skill**, ported from the Codex
one, which is the more mature of the two. Its hard-won details are kept whole —
proving the bridge URL with `/health` before anything else, registering and
writing the credential file in one shell command, the quoted heredoc, the BSD
`printenv` trap, and treating a pasted registration token as already spent.
What differs is host-specific, and mostly warnings: OpenCode's config
interpolates `{env:VAR}` and takes an `environment` key on a server entry, so
there is a plausible-looking place to put the token that would quietly make
every session on the machine the same agent; and it rejects unknown properties
by failing the whole config, for every session, not just yours.

## What this does not do

Switch Console still writes the connector's files itself, and still carries the
embedded copies. It cannot install the published package yet: a pin has to name
a version that exists on the registry, and nothing has been published. That
migration is the follow-up, once the first tag lands.

## Before this can release

- **A trusted-publisher registration on npmjs, made by hand**, before the first
  tag. It is a second registration in its own right — the runtime's does not
  extend to this package — and it names the repository, the workflow filename
  and the environment. None of that is validated when saved, so a mistake
  surfaces at the next release rather than at the mistake.
- **The first publish is a bootstrap**, as the runtime's was.
- No version is set here. Publishing is a tag,
  `switch-connector-opencode-v<version>`.

## Testing

`just artifacts-check` passes; 211 tests in the plugins package pass, including
14 new ones covering the installer against a real config directory — that it
leaves the user's other MCP servers and settings alone, that uninstall takes
back only its own entry, that it is idempotent, and that a stale record with no
server registered reports as not installed rather than as installed.

Not yet run against a live OpenCode session or a real Switch deployment; the
configure skill in particular is where the Codex equivalent found most of its
bugs, and it deserves the same treatment before anyone relies on it.


abeldantas:
Triage note while clearing conflicted PRs: this one's conflicts are semantic, not mechanical, so I left them alone. The branch publishes the connector as an npm package (@sandboxaq/switch-connector-opencode 0.1.2, install-by-name), but main has since moved to the written-by-Console model (renamed package, 0.1.8, nothing fetches it). Whether the published-package approach still applies on top of that is your call — the rebase is a design decision, not a merge.

PR #246, opened 2026-08-17
fix(console): finish the error-message sweep, and a lint rule to hold it (CHOO-2060)

amaudruz:
Second and final PR for [CHOO-2060](https://sandboxquantum.atlassian.net/browse/CHOO-2060). **Stacked on #245** — review that first; this targets it, and I'll retarget to `main` once #245 merges.

**Scope note:** this PR originally also reported background failures that had only ever been logged. That surfaced messages to people who were not being interrupted before, which is a different change from "make the errors users already see better", so it has been removed (`66d8bb1e`). **Nothing here puts a message on screen that was not on screen already** — every message touched was already displayed, just badly.

## 1. The rest of the call sites

Everything remaining goes through `describeFailure` / `failureText` from #245: the room modals, the sidebar and agent menus, remote-host setup, the messaging-app flows, the local and remote server stores, session and location creation.

The pattern replaced is always the same shape — a `Failed to X` stub in the title and the raw exception in the body. Now the title is a sentence about what the user was doing and the diagnostic moves to the description. **Detail is moved, never dropped.**

Along the way: four duplicate local `errorText` / `errorMessage` helpers deleted, and the three copies of the reset-agent toast collapsed into one. The local and remote server stores gained the same headline/detail split `switchServersStore` got in #245, which retires the last of the `Something went wrong` and `Setup failed` banner headings that pushed the real sentence into small print.

## 2. A rule, so it does not come back

`switch-console/no-raw-error-text` bans `e instanceof Error ? e.message : String(e)` in the renderer and points at the shared boundary. It lives in the oxlint plugin the repo already has, scoped by the override mechanism already in `.oxlintrc.json`.

**It earned its place immediately: it found 15 call sites across 12 files that the manual audit had missed** — including a `toast.error`, four store error fields, and the theme-toggle command. All fixed here. The renderer is now clean under the rule, so the sweep is complete **by construction rather than by my assertion** — which is the point of adding it rather than only editing strings.

## Testing

`typecheck`, `lint` (zero errors, including the new rule) and `format` clean. 2,501 node tests and 126 browser tests pass. Two existing tests that asserted the raw error string were updated to the new text, with their intent preserved.

No version numbers touched.

[CHOO-2060]: https://sandboxquantum.atlassian.net/browse/CHOO-2060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amaudruz:
Folded into #245 — one PR, as requested. The crash-screen diagnostics button has been removed too: no new UI, and nothing surfaced that was not surfaced before.

PR #247, opened 2026-08-17
feat(console): send a small set of product events to Amplitude (CHOO-2177)

christian-mcdermott:
## The identifier decision, settled

The consent gate (#215, `CHOO-1955`) has merged and this is rebased onto main.

The payload rule that came with it read *anonymous counters, no identifier of
any kind*, and this PR rewrites it. That is the decision, visible as a diff, and
it has now been taken twice over:

- **By the product side** — per-install identifiers are wanted.
- **By the transport** — the relay's guard drops any payload without a
  canonical-UUID client id, and that id is what becomes Amplitude's device id.
  Sending nothing identifying is not available through this path at all.

What follows from it, and is in this PR: the consent default moves from on to
**off**, the prompt says plainly that an id is sent, and migration `0047` stops
that flip silently revoking an agreement given under the old default.

`CHOO-1370` removed the previous telemetry stack (PostHog) outright. This is a
deliberate reintroduction rather than a revert, and reuses none of that code.

## What it sends

Six events, and no more — the smallest set that answers "is it used, and where
does it break":

- **app_launched**
- **agent_created** — agent type, local or remote
- **session_started** — agent type, local or remote
- **session_ended** — plus `normal` or `failed`
- **server_added** — local, remote managed, or external
- **connector_installed** — agent type, local or remote, success or failure

Every event also carries app version, OS, build channel and the install id.
A question this set cannot answer is the bar for adding to it.

## The decision that changed: consent is now opt-in

Amplitude has no notion of an event without an id, and without one there is no
difference between one user and fifty — so a payload carries a **random
per-install id**. That makes the data pseudonymous rather than anonymous, and
#215's own rule says an opt-out default cannot cover that. So this PR flips the
default from on to off and updates the disclosure copy, the settings row and the
payload rule in `AGENTS.md` to match. Approved as a product decision before
implementation.

The id is random, created on first send, stored locally, and derived from
nothing — not the machine, the user account, the network, or any Switch
identity. It is deliberately **not** the `telemetry:instanceId` the removed
PostHog stack left behind in the databases of installs that predate it, which no
migration ever scrubbed.

## How the promise is kept

- **Consent is read at the point of emission, every time, never cached** — so
  revoking it stops the *next* event, not the next launch. Consent off makes no
  request at all.
- **Nothing free-text can reach a payload.** The event catalogue's property
  types are literal unions and numbers, and — because excess-property checking
  does not apply through a spread — the emitter also drops at runtime any field
  the catalogue does not name. A path, a name, an error message or a prompt
  cannot get through either way. Error *codes* only.
- **The vendors never see the user's address.** They are called by the relay,
  not by the app, so a user's IP cannot be geolocated into an event. This is why
  the consent copy can promise "never your location" as a fact rather than as a
  request — the previous direct-to-Amplitude design could only ask, using a
  sentinel Amplitude does not document.
- **No credential ships in the app.** Events go to the company's public OTLP
  relay, which holds the Amplitude and Datadog keys server-side — the same relay
  a public GitHub Action posts to from a customer's CI. A desktop binary cannot
  keep a secret, so it no longer carries one.
- **A hand-built OTLP record posted with `fetch`, not the OpenTelemetry SDK.**
  What leaves the machine, and when, is the whole promise being made; an SDK
  brings its own batching, retry and identity behaviour. Everything sent is
  visible in one file.
- **Failure never reaches the user**: short timeout, no retry, fire-and-forget.
  A failed send is logged with an `errorCode` and dropped.
- **Inert by default**: no `MAIN_VITE_AMPLITUDE_API_KEY` means it is off and
  says so once. A dev run sends nothing unless `SWITCHDASH_TELEMETRY_DEV=1`.
- **The sidecar reports nothing** — it runs headless with no consent prompt, so
  remote-*started* sessions go uncounted. Stated rather than quietly
  undercounted.

## Verified by watching traffic, not by reading code

Same build, same run, only the consent setting changed:

- Consent **off** — app fully started, **zero requests**; nothing reached the
  listener at all.
- Consent **on** — exactly one event, carrying only the install id, app version,
  OS and build.
- Relaunched — same install id, new event id.

`app_launched` is proven end to end that way. The other five are covered by unit
tests at the point they fire rather than by driving the UI.

Also: `pnpm run typecheck`, `lint`, `format:check`,
`vitest --project node --project main-db` (2693 tests) and
`pnpm run test:migrations` (39) all pass, on the rebase onto current main.

Rebasing brought one behaviour change worth naming: main has since gained the
OpenCode connector, which the app writes rather than installing through a
marketplace CLI. It is still a connector being installed, so it now reports as
one, with tests for both its success and its failure.

## Reviewed, then fixed

A four-pass review of this PR is in the comments, and the second commit acts on
all of it. Three findings were real defects rather than polish: the consent-loss
described above, a remote session pruned from its VM never reporting an end
(two hook buses share the name `session:deleted`, and only one was subscribed
to), and a keyless build being redirectable to any Amplitude project by an
environment variable set before launch.

The review also found the tests weaker than they looked — `trackEvent` mocked
but never asserted, the install id untested entirely, one assertion that could
not fail. That is what most of the new test count is.

## Verified against the real relay

It is public and unauthenticated, so the wire format could be checked rather
than assumed:

- our exact payload → `200 {"partialSuccess":{}}`
- malformed JSON → `400`, so the 200 above means the shape was understood
- a non-UUID client id → `200`, silently dropped. That is why the install id has
  a test pinning it against the relay's own guard expression: a rejected payload
  looks exactly like an accepted one from here.

Then end to end through the app against the live relay: consent off sends
nothing; consent on sends, with no failure logged.

## Nothing further needed to ship

The API key that was outstanding is no longer required — there is no key. The
one open item is a disclosure question, not a blocker: the relay's hostname
belongs to another product and is now written down in a public repository.
Adopted with the collaborating dev's agreement, flagged for the product owner.

## Note for the reviewer

`session:agent-exited` now carries the supervisor's exit decision. That is what
separates a session about to be respawned from one that is over, and without it
a crash-then-recover would be counted as a session ending.


christian-mcdermott:
## Review — four passes (correctness, privacy, conventions, tests/claims)

Self-review, so weighted towards what is wrong. Every finding below was
re-verified against the code by hand; two things the reviewers flagged turned
out to be overstated and are corrected rather than repeated. **Verdict: three
blockers, none of them in the happy path, all cheap to fix.**

---

### Blocking

**1. The default flip silently discards an existing "yes" — a release-ordering
hazard, not a code bug.** `SettingsStore.update()` stores a *delta* against the
defaults (`computeDelta`, `settings/utils.ts:24-35`): a field equal to the
default at write time is not persisted at all. So a user who consented while
`enabled` still defaulted to `true` has only `{askedAt}` in their row. Flip the
default to `false` and that row now resolves to *declined* — and
`TelemetryConsentGate` never asks again, because `askedAt` is set. It fails
closed (nobody is silently opted **in**), but a real "yes" is lost, and it looks
identical to a decline.

Dormant today: neither consent commit has shipped, so no such row exists
anywhere. It goes live the moment #215 ships in one release and this in a
later one. **Either these land in the same release — and someone says so out
loud — or this PR needs a migration that materialises `enabled` into any
existing `telemetry` row before the default moves.** Ordering is the whole
subject of this PR; it should not have a second ordering trap inside it.

**2. `session_ended` never fires for sessions the remote reconciler removes.**
There are two hook buses that both use the name `session:deleted`:
`sessionService`'s lifecycle hooks, and the separate `sessionHooks` singleton,
which is what `remote-session-reconciler.ts:168` emits on when it prunes a
session that has vanished from a VM (or was terminated from another client).
`telemetry-listeners.ts:74` subscribes to the first only.
`search-service.ts:108-111` subscribes to **both**, with a comment explaining
exactly this — the precedent was there and I missed it.

Effect: an adopted remote session reports `session_started` and then nothing,
forever, and its `liveSessions` entry leaks. Remote sessions are precisely the
population this data is meant to say something about.

**3. The API key's `process.env` fallback is not dev-gated.**
`config.ts:59-64` reads `import.meta.env.MAIN_VITE_AMPLITUDE_API_KEY ??
process.env.MAIN_VITE_AMPLITUDE_API_KEY`, unconditionally. The endpoint override
one line below is correctly gated on `build === 'dev'`; this is not. So a
packaged build with no key baked in — which the code documents as inert — will
happily take a key from the environment at launch and send a consenting user's
events to whatever Amplitude project that key belongs to. The fallback exists
only to make the tests work under Vitest. Gate it the same way as its
neighbour.

---

### Should fix here

**4. After any app restart, sessions end as `unknown`.**
`restoreSwitchRoomSessions()` calls `provisionSession()`, which fires
`session:runtime-ready`, not `session:created` — so restored sessions are not in
`liveSessions`, and their eventual end reports `agent_type: 'unknown'`,
`location: 'unknown'`. Starts and ends will not reconcile for anyone who
restarts the app with live sessions, which is everyone. Populating the map on
`session:runtime-ready` as well (without emitting a second `session_started`)
fixes it.

*Corrected from the review:* a restored session that **crashes** is still
reported as `failed` — the exit hook carries the decision regardless of whether
the map knows the session. Only the labelling degrades, not the outcome.

**5. `ip: '0.0.0.0'` is a community convention, not a documented Amplitude
feature.** Amplitude's HTTP V2 docs describe `ip` only as the address to
reverse-lookup a location from, and name exactly one supported way to stop
that: ask Amplitude to drop IP and location at ingestion. The sentinel most
likely works — no GeoIP database resolves a reserved address — but "most
likely" is carrying a promise we make to users in plain words. Keep the field,
and either get the ingestion-side drop configured or verify against really
ingested events via the Export API before trusting the copy.

**6. Three of the six call sites trust the type system where the other two
don't.** `connector_installed` runs `isValidProviderId()` before emitting;
`agent_created`, `session_started` and `session_ended` pass `providerId`
straight through from the database, where the column is a Drizzle `.$type<>()`
cast with no CHECK constraint, reachable via a `createAgent` RPC that does a raw
insert with no validation. Make the three consistent with the two.

**7. "Nothing free-text can reach a payload, by construction" is overstated.**
Excess-property checking does not apply through a spread, and
`telemetry-listeners.ts:44` (`{ ...shape, outcome }`) and
`amplitude-client.ts:56` (`{ ...properties, build }`) are both spreads. Today the
shapes are exactly right, so nothing leaks — but the guarantee is convention,
not the compiler, at those two points. Either narrow the claim or add a runtime
allowlist in `buildAmplitudeEvent`.

---

### Test gaps

- `install-id.ts` — **no test at all**, and the existence of that id is the
  entire reason consent became opt-in. Concurrency dedup, first write, and the
  reset-on-failed-write are all unexercised.
- `switch-setup-service.test.ts` mocks `trackEvent` and never asserts it, while
  calling `install()` three times: coverage that looks real and is not.
  `RemoteSwitchSetupService.install()` has no test whatsoever.
- `servers-store.db.test.ts` runs the *real* `trackEvent` against a real DB
  fixture. Harmless only because the test environment happens to have no API
  key — an accident, not a boundary.
- Untested: the "report the disabled reason once" guard (delete it and every
  test still passes), the config matrix beyond the dev lane, the `timeout`
  error branch, and the insert-only rule for `server_added` — the last being
  exactly what a future refactor would break silently.
- "carries no machine, user or network identity" is theatre: it greps for
  substrings that no current code path could produce. It cannot fail.

### Minor

- `connector_installed` reports a **failure** for an agent type whose install
  was never attempted (the unsupported-type early return). Unreachable today —
  both UI paths pre-filter — but the guard encodes the wrong invariant.
- `endedSessions` grows for the life of the process.
- `time` is stamped after two `await`s rather than when the event happened.
- `telemetry` is missing from the domain list in
  `agents/architecture/main-process.md`.
- The consent copy does not mention that events carry the agent type and
  local-vs-remote, which four of the six do. AGENTS.md requires the copy to
  stay in step with the payload rule, so this is a real gap — though what to
  disclose is a product call.
- The outbound request sets no `User-Agent`; Amplitude parses that header into
  device properties server-side. Low impact, but it is a channel this file
  claims to fully account for.
- **The CHANGELOG entry from #215 is now false** on all three counts (the app
  does send, the toggle defaults off, the payload does carry an identifier). I
  am not permitted to edit `CHANGELOG.md`, so flagging it: it must not ship
  as-is from a public repo.
- PR body: "physically cannot merge ahead of #215" — overstated. The base can
  be retargeted by hand. It is a strong guard rail, not a law.

### Verified sound

The consent gate is read per event and never cached; a throwing settings read
fails closed; nothing is constructed — not even the install id — before the gate
passes; startup order puts the first event strictly after settings
initialisation; the key never reaches the renderer bundle, the logs, or the
diagnostic export; the dev endpoint override genuinely cannot activate in a
packaged build; the sidecar imports no telemetry; a deliberate Stop is never
misreported as a crash; the double-end guard works; `AbortSignal.timeout` really
does reject with `TimeoutError`; no retry; no version bump required; 2334 tests,
typecheck and lint all genuinely pass.

christian-mcdermott:
## Second review — and the fixes for it

Four passes again, weighted at the code the *first* round produced, since none
of that had been reviewed. This time one of them built the app for real and one
went at the migration empirically rather than by reading it. Everything below
was re-verified by hand before being acted on; two findings were overstated and
are recorded as such rather than repeated.

**Four real defects, all now fixed in `1182d8d5`.**

### 1. A session that dies immediately was reported as starting after it ended

`createSession` inserts the row, provisions the runtime and starts the agent
*before* `session:created` fires. The handler then reads the database twice to
learn what the session is — and an agent that fails at once (bad binary, missing
dependency, auth) reports its failure through a synchronous callback while those
reads are still in flight. `session_ended` went out first with `unknown`, then
`session_started` with a later timestamp, and the map entry the created handler
wrote was never removed, because the end that would have cleared it had already
been counted.

Both handlers now check before they write, and an ending clears the entry
whether or not it is the one that reports. The test builds the interleaving
deliberately by suspending the lookup mid-handler — the existing harness awaited
every handler to completion, so the race could not occur in it. Removing the
guard makes that test fail with two events instead of one; I checked.

### 2. The migration could have stopped the app starting

SQLite's JSON functions raise on malformed input, migrations run in one
transaction, and a failed migration is `dialog.showErrorBox` and `app.quit()` —
on every launch, unrecoverably. Nothing writes an unreadable settings row today
(every writer goes through `JSON.stringify`), but the column has no constraint
saying so, and every *reader* already tolerates it. `json_valid(value)` now
guards it, so one unreadable row costs its own repair rather than the whole
application.

Its sibling choice is now pinned by a test too: `json_type` rather than
`json_extract`, because extract cannot distinguish an absent `enabled` from one
explicitly set to `null` and would answer on the user's behalf.

### 3. A test pair that passed only in the order it was written

The service is a module singleton that reports each reason for silence once, so
the "says it once" test depended on the test above it having already consumed
the reason — reproducibly failing under `--sequence.shuffle`. Each takes a fresh
instance now, and "once" is asserted by counting rather than by absence. Six
shuffled runs, clean.

### 4. The consent dialog promised something I could not verify

It listed "your location" as never shared. Nothing sends one and the payload
asks Amplitude not to derive one — but that request is a convention rather than
something Amplitude documents, which the code comment says plainly while the
dialog said otherwise. The promise is out until the ingestion-side setting is
confirmed with Amplitude, at which point it goes back.

### Smaller things fixed

- An event may no longer declare a `build` property, which the emitter adds last
  and would silently overwrite. Adding the check found a real instance:
  `Record<string, never>` for an event with no properties has a `keyof` of
  `string`, so it claimed every ambient name.
- The timestamp fix from the last round had no test — reverting it broke
  nothing. It has one now.
- `agents/architecture/main-process.md` said the listeners file subscribes all
  six events. It subscribes three; the other three have no bus and call
  `trackEvent` at their own site.

### Overstated, and not acted on

- *"`liveSessions` grows unbounded"* — only via the race above, which is fixed;
  entries are removed on every ending, including the second one.
- *"The redundant read on the auto-spawn path"* — real but small: two reads on a
  path that is already opening a terminal. The docstring no longer claims the
  guard always avoids them.

### Verified this round, by doing rather than reading

- **The app builds.** `pnpm run build` clean; the only warnings are two
  pre-existing dynamic-import notices in unrelated files.
- **The key never reaches the renderer.** Built with a key: it appears exactly
  once, in `out/main/index.js`. `grep` across all 104 renderer assets: nothing.
  Built without one, `import.meta.env.DEV` is folded away so hard that
  `telemetryBuildChannel()` cannot return `'dev'` at all — the build is inert as
  claimed, structurally.
- **The migration**, against the real `better-sqlite3` the app ships: produces a
  JSON boolean rather than `1`, preserves the other keys, survives replay,
  no-ops on a database with no telemetry row, and leaves every already-explicit
  row untouched. Journal and snapshot chain correctly to 0046.
- **The delta store** keeps the repaired row rather than collapsing it on the
  next write, and the prompt does not reappear for it.
- **Gate:** typecheck, lint, `format:check`, 2388 tests, 39 migration tests. The
  one lint warning is pre-existing in an unrelated file.

christian-mcdermott:
## Third review — the transport rewrite, and the fixes for it

Three passes again, weighted at the code no one had reviewed: the rebase onto a
much-newer main, and the swap from a keyed Amplitude call to a keyless OTLP post.
One pass read the collector's own source in the other repo and probed the live
endpoint. Everything below was re-verified by hand before being acted on, and
three findings are recorded as *not* acted on, with reasons.

**Five real defects, all fixed in `a2288cc1`.**

### 1. A user-facing promise that depended on another repository's configuration

The consent copy said your location is never shared, on the grounds that the
relay calls the analytics services, so they see its address and not yours. That
is true today — confirmed in the collector's `config.yaml` (no `include_metadata`)
and its registered components (nothing IP- or geo-capable is even compiled in),
rather than from its README.

But it is a property of a repo this one does not own, and that README documents
the exact change that would undo it: read `CF-Connecting-IP` and put the real
client address on the resource, so Amplitude's geography becomes useful. A
reasonable thing for them to ship. The promise here would have turned false
without a line changing in this repo, and no test would have noticed.

It now promises what this app does — sends no address and no location — which is
ours to keep. `AGENTS.md` records why it is phrased that way, so nobody
"improves" it back.

### 2. A misconfigured connector took the whole install down

A plugin can declare a file-based connector and implement no behaviour for it.
Resolving that throws, and the call sat outside the try that turns failures into
results — so the user's install rejected rather than failing: a stack trace in
the UI, and no record that an install was even attempted. Fixed in both the
local and the remote service.

The remote service's file-based install now has tests at all, which it did not:
it was emitting `connector_installed` with no coverage while its local twin had
a suite. Two implementations of one behaviour, and only one was checked.

### 3. A missing property would have been sent as the string `"undefined"`

The catalogue's types cannot be violated through an object literal, but they can
through a spread — which is the hole the runtime allowlist exists to close — and
the allowlist coerced whatever it found. An absent value became the word
"undefined", shipped as though it were an answer. It now refuses: the event is
dropped and the mismatch logged.

### 4. Every event was a blank line in the log sink

The relay forwards the same record to Datadog, which reads the OTLP body as the
message. We sent no body, so events arrived searchable by attribute and empty to
read. It now carries the event name, which is already being sent as an attribute
— nothing new is disclosed.

### 5. The partial-success response was assumed rather than read

OTLP lets a 200 carry a count of records the receiver would not take, and this
is the only channel by which a rejection can ever reach the client: the relay
answers 200 whether it kept the payload or dropped it. Now read and reported.
Worth being precise about the limit — today's relay drops on a failed guard
without populating it, so this closes the reporting gap, not the silent-drop
one.

Also: the request names itself again (the transport swap dropped the explicit
`User-Agent`), and two documents still described a build-time API key and an
"Amplitude sender".

### Found, and deliberately not fixed

- **The files-kind connector's `update()`, `uninstall()` and update-detection
  have no tests.** Real, but it is main's code and main's gap — this PR only
  added telemetry around `install`. Widening the diff to cover it would hide
  the change under unrelated work. Worth its own ticket.
- **Every event is severity INFO**, including a failed session, so Datadog
  cannot alert on severity alone. `outcome` is a facet; this is a preference,
  not a defect.
- **No per-event dedup id.** The relay derives one from a hash of the content,
  so replays already collapse; adding ours would duplicate it.

### Verified this round

- The collector accepts our exact payload, body and all: re-probed live after
  the change → 200. Malformed JSON → 400, so a 200 means the shape was
  understood; a bad client id → 200 and a silent drop, which is why the install
  id is pinned against the guard's own regex.
- The name resolves: `event.name` is second in the exporter's lookup, and the
  filter that drops unnamed records mirrors the same lookup — checked in its
  source, not inferred.
- Attribute counts, value sizes and payload size are far inside every cap the
  relay enforces.
- The build is clean, the endpoint appears only in the main bundle, and nothing
  telemetry-related beyond consent copy reaches the renderer.
- Gate: typecheck, lint, `format:check`, 2698 tests, 39 migration tests, and the
  telemetry suite shuffled four times.

christian-mcdermott:
## How to verify this by hand

Fifteen minutes end to end. Nothing here touches your real install: every step
runs against a scratch database, and the traffic goes to a listener on your own
machine until the last step, where you choose to send it onward.

### 0. Set up

```bash
cd <worktree>/console
pnpm install
pnpm -r --filter './packages/**' run build   # the app build needs these first
```

Open a second terminal and start something to catch the traffic:

```bash
node -e "require('http').createServer((q,s)=>{let b='';q.on('data',c=>b+=c);q.on('end',()=>{console.log('\n--- '+new Date().toISOString());console.log(JSON.stringify(JSON.parse(b),null,2));s.writeHead(200,{'Content-Type':'application/json'});s.end('{\"partialSuccess\":{}}')})}).listen(9009,'127.0.0.1',()=>console.log('listening on 9009'))"
```

Then, in the first terminal:

```bash
cd apps/switch-console-desktop
SWITCHDASH_DB_FILE=/tmp/telemetry-check.db \
SWITCHDASH_TELEMETRY_DEV=1 \
SWITCHDASH_TELEMETRY_ENDPOINT=http://127.0.0.1:9009 \
pnpm run dev
```

Both variables are dev-only and cannot do anything in a packaged build. Without
`SWITCHDASH_TELEMETRY_DEV=1` a dev run sends nothing at all, which is the point
of step 5.

### 1. The prompt appears, and starts off

On a fresh database the first-run prompt should appear, listing what is and is
not shared. **The toggle starts off.** Leave it off and press Continue.

### 2. Consent off means nothing is sent — this is the acceptance criterion

Use the app for a minute: create an agent, start a session, open Settings. The
listener terminal should stay **completely silent**. Not "an empty request" —
nothing at all.

### 3. Turn it on, and watch what actually leaves

**Settings → General → Share usage data**, on. Then quit and relaunch the app
(same command). The listener should print one record. Read it — everything the
app will ever send is in that shape:

- `flint.client_id` — a random UUID for this install. Not derived from anything.
- `service.name` / `service.version` / `os.type` / `os.version`
- `event.name`: `switch_console.app_launched`, and `build: dev`

Check what is **not** there: no paths, no agent or room or project names, no
user or machine name, no IP, no prompt text.

### 4. Trigger the rest of the six

Each should produce exactly one record:

| Do this | Expect |
|---|---|
| Relaunch the app | `switch_console.app_launched` |
| Create an agent | `switch_console.agent_created` |
| Start a session | `switch_console.session_started` |
| Delete or archive that session | `switch_console.session_ended`, `outcome: normal` |
| Add a server (Settings → Servers) | `switch_console.server_added` |
| Install a connector (Settings → Agents) | `switch_console.connector_installed` |

`session_ended` with `outcome: failed` needs an agent that dies and cannot be
recovered — worth skipping unless you want to force it.

Two absences are correct, not bugs: a session still running when you quit
reports nothing, and sessions the remote sidecar starts while the app is closed
are never counted (it has no consent prompt out there).

### 5. Revoking works immediately

Turn **Share usage data** back off, then create an agent. Nothing should reach
the listener. The gate is read per event and never cached, so it takes effect on
the next event rather than the next launch.

Same check for the dev guard: stop the app, drop `SWITCHDASH_TELEMETRY_DEV=1`,
relaunch with consent still on. Silence, and one line in the log saying
`telemetry: not sending … dev_build`.

### 6. Optional — send it for real

Drop `SWITCHDASH_TELEMETRY_ENDPOINT` and the app posts to the real relay
(`telemetry.flintai.dev`), which needs no key. With consent on, relaunch.

**Check Datadog first — it is the control.** The relay runs two pipelines from
the same record: one to Datadog, one to Amplitude. Datadog's has fewer steps and
no dependency on the Amplitude key or project. Look for `service:switch-console`;
the log message is the event name.

- **In Datadog but not Amplitude** → the Amplitude leg. Nothing on this side can
  see it: the relay's Amplitude exporter queues into an SDK buffer and reports
  success regardless, so a rejection appears only in the relay's own logs. That
  is a question for whoever runs the relay, not a change here.
- **In neither** → the record did not survive ingest, or the search is looking
  in the wrong place. Say so and it is worth digging.

**They land in the `flintai` *production* project, not the dev one.** The relay
holds a single Amplitude key, so everything through it — every client, every
build channel — arrives in that one project. Confirmed by observation, after a
long hunt through the wrong project: if you cannot see your events, check which
project you are in before assuming they were lost. `build = dev` and the
`switch_console.` prefix are what separate this app's test traffic from
everything else living there.

**Do not use Amplitude's User Look-Up.** It matches on `user_id`, and this app
deliberately sends only a device id — the relay's own documentation calls out
that a device-only client cannot be found that way. Build a segmentation chart
on the event type `switch_console.app_launched` instead, over today, with no
other filters. A brand-new event type can take a while to appear in Amplitude's
pickers; type the name in by hand rather than waiting for it in the list.

Filter on the event property `build = dev` to separate your test traffic from
real numbers.

### 7. Optional — the consent migration

The one piece with no visible surface. It protects anyone who agreed while the
default was still on, whose "yes" was stored as an absence.

An earlier version of this guide said to "use a fresh database", which cannot
work: on a fresh database the migration runs before there is any row to repair.
The migration has to find the row already there, so rewind the ledger instead.

With the app **not running**, and using the scratch database from the steps
above:

```bash
sqlite3 /tmp/telemetry-check.db   "delete from __drizzle_migrations where created_at >= 1786983032561;
   insert into app_settings (key,value,updated_at)
     values ('telemetry','{\"askedAt\":1700000000000}',0)
     on conflict(key) do update set value=excluded.value;"

sqlite3 /tmp/telemetry-check.db "select value from app_settings where key='telemetry';"
# {"askedAt":1700000000000}     ← an agreement, stored as the absence of a refusal
```

Relaunch the app, quit, and look again:

```bash
sqlite3 /tmp/telemetry-check.db "select value from app_settings where key='telemetry';"
# {"askedAt":1700000000000,"enabled":true}   ← written down explicitly
```

Without the migration that row would read as a refusal under the new default,
and the prompt would never ask again. (Verified by running exactly these
commands.)

### 8. Verify our half without Amplitude, Datadog, or anyone's permission

This runs the relay's *own* guard and name filters in a collector on your
machine, and shows you the record the app produces arriving at the far side of
them. It is the strongest check available from here: everything past this point
is someone else's infrastructure.

```bash
mkdir -p /tmp/otelchk && cat > /tmp/otelchk/config.yaml <<'YAML'
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4318
processors:
  filter/guard:
    error_mode: ignore
    logs:
      resource:
        - 'attributes["flint.client_id"] == nil'
        - 'not IsString(attributes["flint.client_id"])'
        - 'not IsMatch(attributes["flint.client_id"], "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")'
        - 'attributes["flint.client_id"] == "00000000-0000-0000-0000-000000000000"'
      log_record:
        - 'Len(attributes) > 128'
  filter/named_events:
    error_mode: ignore
    logs:
      log_record:
        - 'log.event_name == "" and attributes["event.name"] == nil and attributes["event_type"] == nil'
exporters:
  debug:
    verbosity: detailed
service:
  pipelines:
    logs/amplitude:
      receivers: [otlp]
      processors: [filter/guard, filter/named_events]
      exporters: [debug]
YAML

docker run -d --name otelchk -p 4318:4318 \
  -v /tmp/otelchk/config.yaml:/etc/otelcol-contrib/config.yaml \
  otel/opentelemetry-collector-contrib:latest
```

Run the app against it — note the full path, this endpoint is not the bare host:

```bash
SWITCHDASH_DB_FILE=/tmp/telemetry-check.db \
SWITCHDASH_TELEMETRY_DEV=1 \
SWITCHDASH_TELEMETRY_ENDPOINT=http://127.0.0.1:4318/v1/logs \
pnpm run dev
```

With consent on, `docker logs otelchk` should show, for each event:

```
EventName: switch_console.app_launched
Body: Str(switch_console.app_launched)
     -> service.name: Str(switch-console)
     -> flint.client_id: Str(<a uuid>)
     -> event.name: Str(switch_console.app_launched)
     -> build: Str(dev)
```

**`EventName` on that first line is the thing to check.** It is what the relay's
Amplitude exporter reads first, and a record that reaches it without one is
discarded in silence — the bug this guide's step 6 originally exposed.

Tear down with `docker rm -f otelchk`.

### Clean up

```bash
rm -f /tmp/telemetry-check.db*
```

amaudruz (approved):

christian-mcdermott:
## Prepared for merge — rebased onto main, one real fix

Rebased from `3416620f` onto `5e7a6f6d`, picking up 23 commits including three
console releases (0.27.3, 0.27.4, 0.28.0). Force-pushed; the approval was not
dismissed (`dismiss_stale_reviews_on_push` is off), so **please glance at the
one commit added below** rather than assuming the review still covers
everything.

### Git said clean. It wasn't.

Eight commits replayed with zero conflicts — and the changelog entry still came
out wrong. It was written when 0.27.2 was the newest release, and `git` matched
its hunk on surrounding context: the entry landed **inside `[0.27.3]`**, a
version that shipped on 19 Aug without a line of this code. That section then
had two `#### Added` and two `#### Changed` headings, so a released version was
claiming this feature twice.

`7086a4d3` moves it to `[Unreleased]` and merges it into the headings already
there. The prose said "the previous release described a payload of anonymous
counters" — true of 0.27.2 when written, and three releases stale now — so it
names the version instead.

Worth stating plainly: a conflict-free rebase is not a correct one, and nothing
in the toolchain would have caught this. `[0.27.2]`'s own entry is untouched; it
is accurate for what shipped.

### Verified after the rebase, not before

| | |
|---|---|
| Typecheck | pass, all 5 projects |
| Tests | 2987 desktop (311 files), 202 plugins, incl. 39 migration tests |
| Lint / format | pass |
| CI | all required checks green — `Detect changed trees`, `Secret scan (gitleaks)`, `Switch Console`, semgrep, Socket, CLA |

`mergeStateStatus: CLEAN`.

### Two things checked by hand, because a green build would not have caught them

Main moved through a large console release while this sat. A rebase that
compiles does not prove the emitters still sit on the live paths.

- **The six moments still fire.** Five of them subscribe to typed hook buses
  rather than call sites, so a rename breaks the build — but a *new* code path
  would not. Re-walked them: `agent:created` is emitted from all four agent
  creation paths (`add-agent`, `onboard-agent`, `onboard-location-agents`,
  `attach-configured-agents`), and the session events from `session-service`
  and `session-hooks` as before. The two direct `trackEvent` sites —
  `server_added`, `connector_installed` — are intact in `servers-store.ts`,
  `switch-setup-service.ts` and `remote-switch-setup.ts`, all of which main
  edited during the gap.
- **No migration collision.** Main's latest is `0046`; ours stays `0047`, and
  the journal appends cleanly.

### One pre-existing failure, and it is not this PR's

`src/sidecar/session-spawner.test.ts › trusts the working directory on this VM
before spawning into it` fails locally on macOS. I checked out `origin/main`
into a clean worktree and it fails there identically, so it is inherited, not
caused. It is a macOS-only artifact — the test builds a cwd under `tmpdir()`,
which on macOS is a symlink into `/private/var`, and the trust entry is written
under the resolved path, so the lookup by the unresolved one misses. CI is Linux
and green. Flagged rather than fixed: it belongs to main, and widening this diff
to chase it would bury the change under review.

Also inherited: one `await-thenable` lint warning in `sidecar-runtime.test.ts`,
a file this PR does not touch.

No version numbers were bumped and no manifest touched — that is the releaser's.

PR #281, opened 2026-08-21
renovate-security: scope ticket dedup by label, not by project

paul-abb:
## What

Scope the standing-ticket search in `.github/renovate-security/renovate_jira.py` by label instead of by project.

## Why

The scanner opens one standing ticket per (repo, engine, directory) and refreshes it on each scheduled run. The search that finds that ticket was scoped to a fixed project list. Tickets do not always stay in the project they are created in, so the search could miss the ticket it created on the previous run. The run then opened a second ticket for the same finding.

Without this change, a scheduled scan can open a new ticket for a finding that is already tracked.

## How

- Search by the labels this tool writes at create time, rather than by project.
- Order the results oldest first, so a refresh always updates the first ticket opened for a finding.
- Match the per-directory dimension on an exact summary, which is reliable where a substring search is not.
- Run one search per invocation instead of one per directory.

## Testing

Built the old and the new query and ran both read-only. The old query returned no ticket. The new query returned the open ticket for this repository.

## Screenshots

Not applicable.

## Checklist

- [x] One concern per pull request.
- [x] No change to `jira_client.py`.
- [ ] CLA signed.

### DIFF
--- a/.github/renovate-security/renovate_jira.py
+++ b/.github/renovate-security/renovate_jira.py
@@ -132,20 +132,45 @@ def _tracking_section(
     return f"\n\n## Vulnerability tracking (VULNMGMT)\n\n{lines}\n{marker}\n"
 
 
+_BU_PROJECTS = ("AQMQA", "ALGE", "CAQG", "CHOO", "INFOSEC", "QNV", "SAIGE")
+
+
 def _dedup_projects(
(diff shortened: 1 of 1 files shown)


paul-abb:
Closing this pull request. This repository does not need this workflow.

PR #289, opened 2026-08-24
fix(deploy): the setup hook cannot authenticate on an HTTPS deployment (CHOO-2067)

christian-mcdermott:
The `post-upgrade` setup hook fails on **every deployment served over HTTPS**, and has been worked around with `--no-hooks` rather than diagnosed. Found while deploying switch-dev, where revision 71 failed on it and the two revisions after it skipped it.

## Why it fails

Such a deployment sets `GATEWAY_COOKIE_SECURE`, correctly, so the gateway marks the `switch_auth` session cookie `Secure`. The setup job runs inside the cluster and talks to `http://<service>:8000`.

The interesting part is that a cookie jar **accepts** a Secure cookie over http — it just never sends one back on a non-https request. So the login looks perfect: `200`, cookie visibly in the jar. Every authorized call after it goes out bare and comes back `401`, with nothing naming the cause. That asymmetry is why this presented as an auth failure deep in the script rather than as a login error, and it has its own test.

The fix re-sets the cookie without the flag. Same token; only the rule about which schemes it may travel on is set aside, on a connection that never leaves the cluster.

## The second defect, which the first one exposed

```python
resp = client.get("/gateway/collaborations")
if resp.is_success:
    ...  # return the existing bridge
# falls through
client.post("/gateway/collaborations", json={..., "set_as_default": True})
```

Any failure on that read — a 401, a timeout, a 503 — reads as *"there are no bridges"*, and the script goes on to register a **second** Mattermost bridge and make it the default, taking that from whatever held it. On switch-dev it only stopped because the create was refused as well. Had the read alone failed, the deployment would have quietly gained a duplicate bridge and a new default.

A read that fails is not an empty list. It raises now, and the caller can retry — which a duplicated bridge cannot be undone by.

## Testing

Seven tests, including one that pins the cookie behaviour itself so the premise can't silently stop being true, and one that asserts a 401 on the read posts nothing.

`1775 passed` on the full suite; lint and types clean.

## Note

Reviewed against `main`. The switch-dev deployment currently runs `--no-hooks`; once this lands, the hook can be left on again.

### DIFF
--- a/core/tests/switch_core/deploy/test_setup_gateway_login.py
+++ b/core/tests/switch_core/deploy/test_setup_gateway_login.py
@@ -1,0 +1,180 @@
+"""The bundled setup script has to authenticate over plain HTTP in-cluster.
+
+A deployment served over HTTPS sets `GATEWAY_COOKIE_SECURE`, so the gateway
+marks the `switch_auth` cookie `Secure`. The setup job runs inside the cluster
+and talks to `http://<service>:8000` — where a cookie jar will accept a Secure
+cookie but never send one back. So login returned 200 with the cookie sitting
+visibly in the jar, and every authorized call after it went out bare and came
+back 401. That is what made the post-upgrade hook fail on every HTTPS
+deployment while passing locally.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import os
+import sys
+from pathlib import Path
+from typing import Any
+
+import httpx
+import pytest
+
+_REPO_ROOT = Path(__file__).resolve().parents[4]
+_SETUP_PY = _REPO_ROOT / "deploy" / "shared_resources" / "setup.py"
+
+# setup.py reads its configuration at import time, by design — it is a
+# single-shot container entrypoint, not a library.
+_ENV = {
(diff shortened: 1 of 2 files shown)


christian-mcdermott:
Closing — superseded.

The primary fix here, the Secure-cookie authentication failure, landed independently on \`main\` in 4daa6608594cf58ab12fcbb99891dc2f7da35234 (`fix(deploy): keep the setup job authenticated to the gateway`). Same diagnosis; a different and better remedy — it replays the token as an explicit `Cookie:` header rather than re-setting the cookie in the jar without its `Secure` flag, so the jar is left alone and `Secure` keeps its meaning. The "login succeeded but issued no cookie" guard landed with it.

That leaves only the second defect: `register_bridge` still reads a failed `GET /gateway/collaborations` as "there are no bridges" and falls through to registering a duplicate Mattermost bridge with `set_as_default`. Still present on `main`, and still worth fixing — but this branch is 278 commits behind, conflicting and not rebaseable, so it is carried forward in a fresh PR rather than merged from here.

PR #290, opened 2026-08-25
Confine agent re-registration to the owner's tenant

jameshoweee:
Agent names are global and any user can mint a registration token, so `register_agent(overwrite=True)` let anyone re-register another user's agent by name: the caller got a live API key for the victim's agent (owner_id left unchanged) and the victim's key was deleted. Cross-tenant takeover + DoS of the owner.

Fix: reject overwrite when the existing agent has a different owner, using the same "already exists" error so ownership isn't leaked. Same-owner re-register (connector restart, server rediscovery) is unchanged. Adds a regression test.

### DIFF
--- a/core/tests/switch_core/bridges/agent/protocol/test_register_overwrite_owner_guard.py
+++ b/core/tests/switch_core/bridges/agent/protocol/test_register_overwrite_owner_guard.py
@@ -71,6 +71,10 @@ def _service(session_factory: async_sessionmaker[AsyncSession]) -> ProtocolServi
     svc.client_lifecycle = _FakeClientLifecycle(session_factory)  # type: ignore[attr-defined]
     svc.collab_lifecycle = _NoBridges()  # type: ignore[attr-defined]
     svc.config = SimpleNamespace(jwt_secret_key="test-secret")  # type: ignore[attr-defined]
+    # Duck-typed: the attribute exists on ProtocolService after the api-key
+    # cache landed on main; a no-op stand-in keeps this test valid on both
+    # sides of that merge.
+    svc.api_key_cache = SimpleNamespace(invalidate_agent=lambda *a, **k: None)  # type: ignore[attr-defined]
     return svc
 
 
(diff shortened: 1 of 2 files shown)


jameshoweee:

jameshoweee:
recheck

amaudruz:
recheck

amaudruz:
recheck

jameshoweee:
recheck

abeldantas (approved):
LGTM

core/tests/switch_core/bridges/agent/protocol/test_register_overwrite_owner_guard.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _service(session_factory: async_sessionmaker[AsyncSession]) -> ProtocolService:
+    svc = object.__new__(ProtocolService)
+    svc.session_factory = session_factory  # type: ignore[attr-defined]
+    svc.agent_store = AgentStore()  # type: ignore[attr-defined]
+    svc.api_key_store = ApiKeyStore()  # type: ignore[attr-defined]
+    svc.client_lifecycle = _FakeClientLifecycle(session_factory)  # type: ignore[attr-defined]
+    svc.collab_lifecycle = _NoBridges()  # type: ignore[attr-defined]
+    svc.config = SimpleNamespace(jwt_secret_key="test-secret")  # type: ignore[attr-defined]
```
abeldantas, line 73: Heads-up from an integration run (this branch merged onto today's main): main added `api_key_cache` to `ProtocolService` after this branch was cut, and `_reregister_agent` now invalidates it — so post-merge, `test_same_owner_overwrite_still_works` fails with `AttributeError: api_key_cache`. One-liner keeps the test valid on both sides:

```suggestion
    svc.config = SimpleNamespace(jwt_secret_key="test-secret")  # type: ignore[attr-defined]
    # Duck-typed: ProtocolService gained api_key_cache on main; a no-op
    # stand-in keeps this test valid on both sides of that merge.
    svc.api_key_cache = SimpleNamespace(invalidate_agent=lambda *a, **k: None)  # type: ignore[attr-defined]
```

Verified: with this, the file passes both on this branch alone and merged onto main (ran the merged combination locally with the whole security batch).

abeldantas: Applied directly to the branch (maintainer edit) so the batch can land today.

PR #309, opened 2026-08-27
feat(console): reap orphaned agent runtimes at boot

amaudruz:
Follow-up to #307. The runtime now exits when its host dies and clears up predecessors on startup — but both reach a machine only through the **connector pin**, which moves when a user updates a marketplace plugin. Someone who never revisits Settings → Agents keeps their backlog indefinitely. This reaches them when they update the app instead.

The populations overlap less than they look:

| | app reaper (this PR) | runtime self-sweep (#307) |
|---|---|---|
| Gated on | app update | publish → pin bump → plugin update |
| Console user who never clicks Update in Settings → Agents | ✅ | ❌ |
| Standalone Codex, no Console | ❌ | ✅ |
| Marketplace-only Claude Code user | ❌ | ✅ |

Neither alone covers the fleet.

## One predicate, not two

`reapOrphanedRuntimes` moved to the runtime package's public surface and the app imports it. Two copies of "is this runtime abandoned?" is a failure mode this repo already documents — the sidecar mirroring the desktop app, `switch_hook.py` and `bin.ts` resolving identity twice — and here a disagreement doesn't cause drift, it kills a live session.

It now **returns** what it did instead of taking a log callback. Two callers, two ideas of a log line: the runtime writes prose to stderr, the app wants an enumerated `event`. Threading either through the sweep was the wrong shape.

```ts
export type ReapOutcome = {
  reaped: number;
  removedSessionDirs: number;
  failures: Array<{ stage: 'scan' | 'sweep'; error: unknown }>;
};
```

## Placement

Unawaited, after the RPC router is up and **before session relaunch** — so a runtime abandoned by a previous run is gone before its replacement starts, without the window waiting on a full process scan and a possible few thousand directory removals.

## Verified against real orphans, twice

Not mocks. Orphans built the way the bug builds them (host killed with its stdin held open by a third party, wrappers confirmed reparented to pid 1), then the app started with an isolated `SWITCHDASH_DB_FILE` so it couldn't relaunch real sessions.

**Positive —** four genuine orphans present at boot:

```
reapOrphanedAgentRuntimes: cleared runtimes left by earlier sessions
  { event: 'agent_runtime_reap', reaped: 4, removedSessionDirs: 3 }
```
All four gone. No errors in the app log.

**Negative —** a *healthy* session running, host alive:

```
  { event: 'agent_runtime_reap', reaped: 0, removedSessionDirs: 2 }
```
Its runtime, its wrapper and its session directory (`~/.switch/sessions/45597`) all untouched.

**The second test matters more than the first.** Killing a live session is far worse than leaving an orphan, and it is the case a unit test can only partly stand in for. The predicate spares anything whose ancestor chain reaches a live non-runtime process; the app also passes `keepSessionDir: null`, since it owns no session directory and guessing one could spare a directory that should be swept.

## Tests

- `reap-orphaned-runtimes.test.ts` — 7 tests on the app wrapper (sweeps the shared root, keeps nothing, silent when there is nothing to report, one event when there is, one warning per failed stage, propagates a hard failure to the caller's `.catch` rather than into boot)
- `reap.test.ts` — 2 added for the nullable `keep`
- Package: 96 passing. App: 222 passing in `src/main/core/agent-runtime/`, 3041 across `node` + `main-db`.

`typecheck` and `lint` clean. Two pre-existing failures on `main` are unaffected and unrelated — `js-yaml` types in `scripts/`, and `session-spawner.test.ts > trusts the working directory on this VM`; both reproduce on a clean tree.

## Not done here

**No version bump**, consistent with #307 and your call. `console/AGENTS.md` asks for one on any runtime change, and this touches `reap.ts`, `bin.ts` and `index.ts` — worth deciding before the runtime release goes out, since the pins must name a published version.

The `switch_hook.py` port-lookup bug from #307 is still open and still unfixed.

### DIFF
--- a/console/apps/switch-console-desktop/src/main/core/agent-runtime/reap-orphaned-runtimes.test.ts
+++ b/console/apps/switch-console-desktop/src/main/core/agent-runtime/reap-orphaned-runtimes.test.ts
@@ -1,0 +1,95 @@
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const sweep = vi.fn();
+const logInfo = vi.fn();
+const logWarn = vi.fn();
+
+vi.mock('@sandboxaq/switch-agent-runtime', () => ({
+  reapOrphanedRuntimes: (...args: unknown[]) => sweep(...args),
+}));
+
+vi.mock('@main/lib/logger', () => ({
+  log: { info: (...a: unknown[]) => logInfo(...a), warn: (...a: unknown[]) => logWarn(...a) },
+}));
+
+const { reapOrphanedAgentRuntimes } = await import('./reap-orphaned-runtimes');
+
+const NOTHING = { reaped: 0, removedSessionDirs: 0, failures: [] };
+
+beforeEach(() => {
+  sweep.mockReset();
+  logInfo.mockReset();
+  logWarn.mockReset();
+  sweep.mockResolvedValue(NOTHING);
+});
+
+afterEach(() => vi.restoreAllMocks());
+
+describe('reapOrphanedAgentRuntimes', () => {
+  it('sweeps the shared session root', async () => {
+    await reapOrphanedAgentRuntimes();
+    expect(sweep).toHaveBeenCalledWith({
+      sessionsRoot: path.join(os.homedir(), '.switch', 'sessions'),
+      keepSessionDir: null,
+    });
+  });
+
+  /**
+   * The app owns no session directory. Passing one of a runtime's — or a
+   * plausible-looking guess — would spare a directory that should be swept, and
+   * on a pid collision could spare a live one's.
+   */
+  it('spares no session directory, having none of its own', async () => {
+    await reapOrphanedAgentRuntimes();
+    expect(sweep.mock.calls[0][0].keepSessionDir).toBeNull();
+  });
+
+  it('says nothing when there was nothing to clear', async () => {
+    await reapOrphanedAgentRuntimes();
+    expect(logInfo).not.toHaveBeenCalled();
+    expect(logWarn).not.toHaveBeenCalled();
+  });
+
+  it('reports what it cleared, under one event', async () => {
+    sweep.mockResolvedValue({ reaped: 3, removedSessionDirs: 12, failures: [] });
+    await reapOrphanedAgentRuntimes();
+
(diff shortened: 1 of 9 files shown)


amaudruz:
## Added: a one-shot connector catch-up

Rebased onto the runtime cascade (#308, #310), and added the piece that closes the delivery gap those PRs leave: a connector's pinned runtime version only moves when a user goes looking for it in Settings → Agents, so someone who never does keeps whatever they first installed, across every release.

`catch-up-connectors.ts` + 11 tests.

**Deliberately not a standing auto-updater.** It latches a generation marker on success and does nothing on every later launch — the Update button stays the normal path. Bumping `CATCH_UP_GENERATION`, a deliberate edit rather than a release, is what makes it run again. Same shape as the agent-storage migration marker.

### What it refuses to do

- **Install a connector that isn't installed.** Which agents have one is the user's decision; this is a version catch-up, not an installer.
- **Latch when the marketplace couldn't be refreshed.** `updateAvailable` is false both for "no update" and "couldn't look" — spending the single shot on an answer that never arrived is the one way a one-shot fails silently.
- **Retry forever.** Three attempts, then it stops. "Runs every launch" is exactly what this is not.

One connector throwing (a missing CLI, say) doesn't abandon the rest; it counts as a failure so the attempt cap still applies.

### Verified against a real gap

My machine happened to be two versions behind, so this was a genuine test rather than a staged one. First launch:

```
connector_catch_up_updated   claude    0.9.8 -> 0.9.10
connector_catch_up_updated   opencode  0.1.5 -> 0.1.6
connector_catch_up_complete  updated: 2
```

Confirmed in the UI as well as the log, and `claude plugin list` now reports `0.9.10`.

Second launch, same install: **no catch-up activity, no marketplace call, nothing logged.** That's the latch working, and it's the property worth reviewing — the failure mode here isn't "doesn't update", it's "quietly becomes a per-launch job".

### On the Codex risk from earlier

Codex's update is remove-then-add (`updateArgs: () => null`), so a failed re-add leaves no connector. That risk is much smaller for a one-shot than for a standing updater, and it self-heals: a failure doesn't latch, so the next launch retries — up to the cap. Worth a reviewer's eye on whether three attempts is the right number.

### Gate

`typecheck` and `lint` clean. 3052 passing across `node` + `main-db`. Two pre-existing issues on `main` are unaffected and reproduce on a clean tree: `js-yaml` types in `scripts/`, and `session-spawner.test.ts > trusts the working directory on this VM`.

PR #313, opened 2026-08-27
Fix !command so @mentions resolve and code-wrapped commands run, like /command (CHOO-2342)

abeldantas:
!command is parsed straight from message text. After CHOO-2277 gave each agent an autocomplete user group, that broke bang commands: a typed @agent arrived as Slack markup, and code-wrapped commands were dropped silently. In this PR we fix both, in the shared command funnel and the Slack adapter.


amaudruz (changes requested):
Image

When commands have to be stripped the result comes in the root of the conv, while it should come in the thread of the command message

abeldantas:
SCR-20260828-tagn

amaudruz (approved):

PR #314, opened 2026-08-27
Tell the user why "Add agent" is disabled, on hover over the button (CHOO-2455)

abeldantas:
The New agent dialog greys out "Add agent" whenever the form is not ready but doesn't say why. In this PR we add a hover tooltip that names the reason.


amaudruz (approved):
Sweet!

PR #316, opened 2026-08-28
fix(slack): normalize typed command mentions

### LINKED ISSUE #315: Typed `!`-commands don't normalise Slack mention encoding in arguments
**Repo:** `sandbox-quantum/switch` @ `d1849c92` (all line numbers are against this commit)
**Component:** collaboration bridge — Slack adapter (Discord likely affected; Mattermost to confirm)
**Severity:** Low–Medium — intermittent, no data loss, but the failure is silent
**Reported by:** safi.amin · **Corrected after code review by:** switch-worker-louis-remote

## Summary

Slack's typed `!command` path passes command arguments to the dispatcher **raw**, while the `/slash-command` path first normalises Slack's `<@U123>` mention encoding. When Slack autocompletes an `@name` in a typed command into a real user/bot mention, the dispatcher receives `<@U123>` instead of `@name`, target resolution fails, and the command appears to do nothing.

This is intermittent by nature — it depends on whether the name the user typed autocompleted — which is why it was originally mis-reported as "`!invite-agent` never works".

## Root cause

Two entry points converge on the same command dispatcher, but only one normalises its arguments.

**Slash path — normalises.** `core/switch_core/bridges/collaboration/slack/adapter.py:891`

```python
# Slack encodes any @mentions in the slash text as `<@U…>`; normalise
# them to `@name` so the command dispatcher's targeting resolves.
args = self.translate_inbound(text)
```

The comment there asserts the typed path already behaves this way. It does not.

**Typed `!` path — does not normalise.** `core/switch_core/bridges/collaboration/slack/adapter.py:826-830`

```python
stripped = text.strip()
if stripped.startswith("!") and self._on_command:
    parts = stripped.split(None, 1)
    command = parts[0].lstrip("!")
    args = parts[1].strip() if len(parts) > 1 else ""
```

`args` goes to `InboundCommand` unmodified. `translate_inbound` (`slack/adapter.py:657`) is what rewrites `<@U…>` → `@name` and `<url|label>` → `[label](url)`.

## Steps to reproduce

1. Slack channel linked to a Switch room, with at least one agent registered.
2. Type `!invite-agent @` and **select the target from Slack's autocomplete popup** (so Slack encodes it as a mention), rather than typing the name out as plain text.
3. Send.

**Expected:** the agent is added to the room.
**Actual:** nothing happens — the dispatcher gets `<@U123>` as its first `@token` and cannot resolve it.

**Contrast:** `/invite-agent @agent-name` with the same autocompleted mention works, because the slash path calls `translate_inbound`. Typing the name as plain text (no autocomplete) also makes the `!` form work — hence the intermittency.

## Suggested fix

Apply `translate_inbound` to `args` in the message path, mirroring the slash path:

```python
args = self.translate_inbound(parts[1].strip()) if len(parts) > 1 else ""
```

Then update the stale comment at `slack/adapter.py:891`, which currently claims parity that does not exist.

## Same asymmetry in other adapters — check in the same pass

- **Discord** — `core/switch_core/bridges/collaboration/discord/adapter.py:706-714`: `args=parts[1].strip() if len(parts) > 1 else ""`, raw. Discord encodes mentions as `<@123456>`. Likely affected.
- **Mattermost** — `core/switch_core/bridges/collaboration/mattermost/adapter.py:1079-1082`: raw. Likely benign (sends `@username` as literal text) — confirm rather than assume.

## Suggested regression test

`core/tests/switch_core/bridges/collaboration/test_slack_adapter.py` — feed a message event with `text = "!invite-agent <@U123>"` and assert the resulting `InboundCommand.args` is the normalised `@name`, not the raw encoding.

---

## Note: this corrects an earlier (unfiled) report

An earlier draft titled *"Slack /invite fails for Switch agents"* was **wrong on three counts**. Recording them so any fix is scoped against the corrected framing above, not the old one:

1. **"`!invite-agent` does NOT work"** — false. It is implemented and wired up (`slack/adapter.py:825` dispatches `!`-prefixed messages; `core/switch_core/bridges/agent/commands.py:1077` registers `invite-agent`). It succeeded in a live channel during the reporting session.
2. **"`/invite @agent-name` fails — bug"** — not a bug. `/invite` is Slack's *native* channel-invite command, which takes a Slack user; a Switch agent is not a Slack user, so the failure is correct behaviour and out of scope.
3. **"Only the slash command version works"** — too strong. Both work; the typed form breaks only when the argument was autocompleted into a mention. That conditional nature is the actual signal pointing at the real cause.

Pumpurri:
### Context

Fixes #315 

Slack encodes an autocomplete-selected mention as `<@U123>`, even though it displays the mention as `@username` in the client.

The Slack slash-command path normalizes this encoding with `translate_inbound`, but the typed `!command` path previously passed its arguments to the command dispatcher unchanged. As a result, commands such as `!invite-agent` could silently fail when their target was selected from Slack’s autocomplete popup.

### Description

Normalize typed Slack `!command` arguments using `translate_inbound` before dispatching them.

This converts Slack-encoded mentions such as `<@U123>` into the `@username` format expected by the command dispatcher. The stale slash-command comment was also updated to describe the shared normalization behavior accurately.

The related adapters were checked in the same pass:

- Discord already normalizes typed-command mentions and has regression coverage.
- Mattermost sends mentions as literal `@username`, so no normalization change is required.

### Steps to review

- `core/switch_core/bridges/collaboration/slack/adapter.py` — confirm typed command arguments pass through `translate_inbound`.
- `core/tests/switch_core/bridges/collaboration/test_slack_adapter.py` — regression test sends `!invite-agent <@U999>` and verifies the dispatcher receives `@worker`.
- Confirm the slash-command path continues using the same inbound translation.

Tests:

- `just check` — passed
- `just typecheck` — passed
- `just test` — 2011 passed, 5 deselected
- Live Slack test — an autocomplete-selected mention in `!invite-agent @username` successfully added the corresponding Switch agent

### DIFF
--- a/core/switch_core/bridges/collaboration/slack/adapter.py
+++ b/core/switch_core/bridges/collaboration/slack/adapter.py
@@ -2009,7 +2009,7 @@ class SlackAdapter(CollaborationAdapter):
         if stripped.startswith("!") and self._on_command:
             parts = stripped.split(None, 1)
             command = parts[0].lstrip("!")
-            args = parts[1].strip() if len(parts) > 1 else ""
+            args = self.translate_inbound(parts[1].strip()) if len(parts) > 1 else ""
             await self._on_command(
                 InboundCommand(
                     channel_id=channel_id,
@@ -2072,9 +2072,9 @@ class SlackAdapter(CollaborationAdapter):
         user = await self._resolve_user_name(user_id)
         channel_name = str(payload.get("channel_name", "")) or None
 
-        # Slack encodes any @mentions in the slash text as `<@U…>`; normalise
-        # them to `@name` so the command dispatcher's targeting (first @token →
-        # target agent/role) resolves the same way it does for a typed command.
+        # Slack encodes @mentions in command arguments as `<@U…>`; normalise
+        # them to `@name` so the command dispatcher's first `@` token resolves
+        # to the target agent or role.
         args = self.translate_inbound(text)
 
         try:
--- a/core/tests/switch_core/bridges/collaboration/test_slack_adapter.py
+++ b/core/tests/switch_core/bridges/collaboration/test_slack_adapter.py
@@ -1028,6 +1028,30 @@ def _capture_commands(adapter: SlackAdapter) -> list[InboundCommand]:
     return captured
 
 
+def test_typed_command_translates_slack_mention() -> None:
+    adapter = _adapter()
+    adapter._user_cache["U1"] = SlackUser(name="jane", display_name="Jane")
+    adapter._user_cache["U999"] = SlackUser(name="worker", display_name="Worker")
+    adapter._channel_name_cache["C123"] = "general"
+    commands = _capture_commands(adapter)
+
+    _run(
+        adapter._handle_message_event(
+            {
+                "channel": "C123",
+                "ts": "100.1",
+                "user": "U1",
+                "text": "!invite-agent <@U999>",
+                "channel_type": "channel",
+            }
+        )
+    )
+
+    assert len(commands) == 1
+    assert commands[0].command == "invite-agent"
+    assert commands[0].args == "@worker"
+
+
 def test_slash_command_maps_to_in_room_command() -> None:
     adapter = _adapter()
     fake = _FakeSlashWebClient()


Pumpurri:

amaudruz (approved):
Nice, thanks for this !

PR #319, opened 2026-08-29
fix(discord): anchor inbound message-map on thread root so directly-created threads keep threading

### LINKED ISSUE #318: Discord: directly-created threads lose threading after the first message (message-map keyed on message ref, not thread root)
## Summary

Discord bridge: inbound messages from a Discord thread lose their thread relation whenever the thread was **not** created from an existing message. Discord's direct "Create Thread" flow gives the thread a fresh channel id unrelated to any message id, and once that happens every reply after the first posts top-level in Matrix, forever.

Log signature:
```
[BRIDGE-IN] no Matrix event mapped for external root <parent>:<thread_id>; posting top-level in room ...
```

## Root cause

`core/switch_core/bridges/collaboration/bridge_core.py`, `_handle_inbound_message` (v0.21.0, ~line 532 and ~583):

```python
await self._record_message_map(
    external_channel_id=msg.channel_id,
    matrix_event_id=event_id,
    external_post_id=msg.message_ref,   # <-- always the message's OWN ref
)
```

The thread lookup, a few lines earlier in the same method (~line 497-498), keys on the *thread root*, not the message:

```python
if msg.root_id is not None:
    thread_root_id = await self._matrix_event_for_external_post(msg.root_id)
```

`msg.root_id` is computed in `discord/adapter.py:1146` as `f"{parent_id}:{thread_channel_id}"`. For a thread created via Discord's "reply in thread on an existing message" flow, `thread_channel_id` happens to equal that message's id, so the two keys coincide by accident. For a thread created via Discord's direct **"Create Thread"** button (not attached to any message), the thread channel id is freshly minted by Discord and never equals any message id — so the recorded key (`{channel}:{message_id}`) and the looked-up key (`{parent}:{thread_channel_id}`) can never match.

The sibling method `_handle_inbound_command` (~line 622) already handles this correctly:

```python
thread_root_post = cmd.root_id or cmd.message_ref
```

`_handle_inbound_message` never adopted that pattern — it always anchors on `msg.message_ref`, dropping `msg.root_id` entirely when recording (even though it reads `msg.root_id` for the lookup).

## Evidence (live v0.21.0 deployment)

`bridge_message_map` table holds rows keyed like:
```
1543280319908548658:1543281108840620254
1543280319908548658:1543281133758718032
```
(`{channel_id}:{message_id}` for each message posted into the thread), while the lookup for the next reply computes:
```
1542465297741512754:1543280319908548658
```
(`{parent_channel_id}:{thread_channel_id}`) — a key that was never recorded and can never match.

## Impact

- Every reply after the first, in a directly-created Discord thread, posts top-level in the Matrix room instead of threading.
- Agents subscribed to the bridge receive these events with no `thread_id`, so they lose the thread context for the whole conversation.
- Silent: the only signal is a warning log line; nothing surfaces to the Discord user.

## Repro

1. In a bridged Discord channel, click **Create Thread** directly on the channel (not "reply in thread" on an existing message) — this mints a thread channel id unrelated to any message.
2. Post a message in the new thread. It bridges fine (first message of a thread has nothing to anchor to yet).
3. Post a second message in the same thread.
4. Observe `[BRIDGE-IN] no Matrix event mapped for external root <parent>:<thread_id>; posting top-level in room ...` in the logs, and the message landing top-level in the Matrix room instead of inside the thread.

## Suggested fix

Anchor the recorded correlation on `msg.root_id or msg.message_ref` (mirroring `_handle_inbound_command`'s `cmd.root_id or cmd.message_ref`), and only record when `thread_root_id is None` (the first bridged message anchors the thread; re-recording afterward would just shadow the already-resolved root). Non-thread messages are unchanged since `msg.root_id` is `None` for them.

I have this fix ready and will open a PR shortly with a regression test.

---
🤖 Filed by eq3 (AI agent) on behalf of Tony (meganechan) — found while self-hosting Switch v0.21.0

meganechan:
Fixes #318

## What

`_handle_inbound_message` in `core/switch_core/bridges/collaboration/bridge_core.py` now anchors the recorded message-map correlation on `msg.root_id or msg.message_ref` instead of always `msg.message_ref`, and only records it when `thread_root_id is None` (i.e. the first bridged message of a thread).

## Why

Discord's direct **"Create Thread"** flow mints a thread whose channel id is unrelated to any message id. The thread *lookup* (a few lines above the change) already keys on `msg.root_id`, but the *record* was keying on the message's own ref — so a directly-created thread's message-map entry could never be found by the next reply's lookup. Every reply after the first in such a thread posted top-level in Matrix, and agents received those events with no `thread_id`.

The sibling method `_handle_inbound_command` already anchors correctly (`cmd.root_id or cmd.message_ref`, line ~622) — this brings `_handle_inbound_message` in line with that existing precedent. Non-thread (top-level) messages are unaffected: `msg.root_id` is `None` for them, so the anchor is unchanged (`msg.message_ref`).

## How verified

- Root-caused against a live self-hosted v0.21.0 deployment: `bridge_message_map` held rows keyed `{channel_id}:{message_id}` for messages posted into a directly-created thread, while the lookup for the next reply computed `{parent_channel_id}:{thread_channel_id}` — a key that was never recorded. Full evidence + repro steps in #318.
- Added `core/tests/switch_core/bridges/collaboration/test_bridge_inbound_thread_anchor.py`:
  - `test_directly_created_thread_second_reply_still_threads` — simulates a thread whose id shares no id with any message (the Discord direct-create case); asserts the anchor recorded after the first reply is the thread id, and the second reply resolves and threads under it.
  - `test_top_level_message_still_anchors_on_its_own_ref` — regression guard that top-level messages are unaffected.
  - Verified the first test **fails** against the pre-fix code (`git stash` the fix, rerun) and passes after — confirms the test actually exercises the bug, not just the happy path.
- `uv run ruff check` clean on both changed files.
- Did not run the full `just test` suite (requires a real PostgreSQL instance per `core/tests/` conventions) or `just typecheck` — happy to run these if a maintainer points me at CI, or CI will confirm on this PR.

## Precedent referenced

`_handle_inbound_command`, same file, ~line 622:
```python
thread_root_post = cmd.root_id or cmd.message_ref
```

---
🤖 Filed by eq3 (AI agent) on behalf of Tony (meganechan) — found while self-hosting Switch v0.21.0

### DIFF
--- a/core/switch_core/bridges/collaboration/bridge_core.py
+++ b/core/switch_core/bridges/collaboration/bridge_core.py
@@ -504,6 +504,13 @@ class BridgeCore:
                     matrix_room_id,
                 )
 
+        # Where later replies should find this conversation: the thread's own
+        # key when this message sits inside a thread (so the NEXT reply in a
+        # directly-created thread resolves), its own ref when top-level (so a
+        # thread created FROM this post resolves). Mirrors
+        # _handle_inbound_command, which already anchors on root_id.
+        thread_anchor_post = msg.root_id or msg.message_ref
+
         # An attachment the platform offered but we could not relay must be
         # visible in the room, not swallowed. Append it to the message body so
         # both the agent and the humans see that a file went missing.
@@ -529,12 +536,16 @@ class BridgeCore:
                     matrix_room_id,
                 )
                 return
-            # Record the correlation so a later reply (either direction) threads.
-            await self._record_message_map(
-                external_channel_id=msg.channel_id,
-                matrix_event_id=event_id,
-                external_post_id=msg.message_ref,
-            )
+            # Record the correlation so a later reply (either direction)
+            # threads. Only the first bridged message of a thread anchors it —
+            # once the anchor resolves, thread_root_id is set and re-recording
+            # would just shadow it.
+            if thread_root_id is None:
+                await self._record_message_map(
+                    external_channel_id=msg.channel_id,
+                    matrix_event_id=event_id,
+                    external_post_id=thread_anchor_post,
+                )
             return
 
         # Caption convention: the text rides as the caption on the first
@@ -580,11 +591,11 @@ class BridgeCore:
             if index == 0:
                 first_event_id = event_id
 
-        if first_event_id is not None:
+        if first_event_id is not None and thread_root_id is None:
             await self._record_message_map(
                 external_channel_id=msg.channel_id,
                 matrix_event_id=first_event_id,
-                external_post_id=msg.message_ref,
+                external_post_id=thread_anchor_post,
             )
 
     async def _handle_inbound_command(self, cmd: InboundCommand) -> None:
--- a/core/tests/switch_core/bridges/collaboration/test_bridge_inbound_thread_anchor.py
+++ b/core/tests/switch_core/bridges/collaboration/test_bridge_inbound_thread_anchor.py
@@ -1,0 +1,132 @@
+"""A directly-created thread must keep threading past its first reply.
+
+Discord's "Create Thread" flow (not "reply in thread" on an existing message)
+mints a thread whose channel id is unrelated to any message id. `msg.root_id`
+then carries that thread id, never a message id — so recording the
+correlation under the message's own ref (as `_handle_inbound_message` used
+to) can never be looked up again by `_matrix_event_for_external_post`, which
+keys on `root_id`. `_handle_inbound_command` already anchors on
+`root_id or message_ref`; these pin the same anchor for the message path.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import Any
+
+from switch_core.bridges.collaboration.bridge_core import BridgeCore
+from switch_core.bridges.collaboration.models import InboundMessage
+
+
+class _FakePuppet:
+    matrix_user_id = "@puppet:s"
+
+    def __init__(self) -> None:
+        self.messages: list[dict[str, Any]] = []
+        self._next_event = 0
+
+    async def send_message(
+        self, matrix_room_id, content, format=None, thread_root_id=None
+    ):  # noqa: ANN001, ANN201, A002
+        self.messages.append({"content": content, "thread_root_id": thread_root_id})
+        event_id = f"$evt-{self._next_event}"
+        self._next_event += 1
+        return event_id
+
+
+class _FakeAdapter:
+    def translate_inbound(self, content: str) -> str:
+        return content
+
+
+def _fake_bridge(*, matrix_events: dict[str, str]) -> SimpleNamespace:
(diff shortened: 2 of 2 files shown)


meganechan:

meganechan:
Ran the full local pre-PR verification (macOS, Python 3.13.11, uv 0.9.18) at HEAD (cc98ef6):

- `ruff format --check` + `ruff check` — pass (366 files formatted, no lint issues; pushed a small reformat commit for the new test file to match)
- `mypy --config-file core/pyproject.toml core/switch_core/ connectors/` — pass, no issues found in 185 source files
- `pytest -c core/pyproject.toml core/tests/` — pass, 2013 passed, 5 deselected (integration-marked), 176.25s. Store tests ran against a real PostgreSQL via testcontainers, not mocks.

All green — closing the local-verification gap noted in the PR description.

— eq3 (AI agent) for @meganechan

amaudruz:
Thanks for this — the write-up in #318 is excellent, and the DB evidence made it quick to confirm. The diagnosis is right: the record keys on the message's own ref, the lookup keys on the thread root, and on Discord those only coincide when the thread was created *from* a message.

I'd like to fix it a bit differently, though — here's why.

The awkward constraint:

```python
UniqueConstraint("bridge_id", "matrix_event_id"),
UniqueConstraint("bridge_id", "external_post_id"),
```

One Matrix event gets exactly one external post id. So each bridged message has a single key, and that key is doing two different jobs:

1. **message identity** — which Discord message is this Matrix event? (the 👀 read marker via `anchor_message_ref`, plus edit/delete sync)
2. **thread anchor** — which Matrix event does this external thread hang off?

On Mattermost and Slack those are the same thing, because a thread root really is a message in the channel. On Discord a thread is a *channel*, not a message, so they come apart. That's the real defect — one column doing two jobs.

Because there's only one key per event, any in-place fix has to pick one job over the other. This PR picks threading: with `if thread_root_id is None`, thread replies stop recording a mapping at all. Two problems with that:

- It's shared code, so it also changes Slack, Mattermost, Telegram and Teams — none of which have the bug.
- `_external_post_for_matrix_event(event.anchor_event_id)` (`bridge_core.py:1600`) then returns `None` for in-thread messages, so the 👀 marker silently stops landing on them, on every platform. Not covered by the tests here.

So rather than trade one for the other, I'd rather split them:

- new `bridge_thread_map(bridge_id, external_thread_id, matrix_thread_root_event_id)`
- threading reads and writes that
- `bridge_message_map` goes back to being purely message ↔ message, unchanged for everyone

That fixes Discord with no trade-off and no behaviour change elsewhere. It needs a migration, so it's a bit more than this diff.

Would you be up for taking that on? Your test file carries over almost as-is — it's exactly the right case to assert. If you'd rather not, no problem at all, we'll pick it up from here and credit the issue.

PR #325, opened 2026-09-01
feat(console): offer a slug instead of rejecting an agent name

raullopez-sandbox:
Stack 1/2 — base for `feat/display-name-01-column`.

Currently, typing `Switch Dev` into the new-agent name field got you an error and nothing else. The user has to work out the rule from the error and retype.

This PR is the first of many that will work over a general solution. In this case, we are displaying the user a small button that proposes a name that fulfills the constraints

image

[new-agent-slug-offer.webm](https://github.com/user-attachments/assets/68145f76-d26f-4722-b542-336305d4d2fc)

Next PRs will also add a new field, so that the user can actually write a "Display name" for the agent, and this name will be used through all the different channels automatically


raullopez-sandbox:

amaudruz (changes requested):
let's have some red popup so that it is more obvious that there is something wrong 

For instance could be the text below explaining what to use or not

raullopez-sandbox:
> let's have some red popup so that it is more obvious that there is something wrong

Done ✅ 
image


I saw the class text-destructive is being used in 20+ places but it doesnt have anything set in CSS, so its black everywhere. I will open another PR to fix that too

amaudruz (approved):

PR #335, opened 2026-09-01
feat(resources): user-defined external reference types

raullopez-sandbox:
Reference types are no longer a closed set of four. Any signed-in user can define one — slug, display name, agent-facing instructions, value hint, and a read/write visibility pair, exactly like a Reference.

## What changes

Types were a static in-process Pydantic registry. They are now DB-backed rows resolved per-principal, with the four built-ins (Google Drive, Confluence, GitHub, Jira) still living in code. Built-ins cannot be edited or deleted and **win any slug collision**, so no user type can shadow one.

- **New `reference_types` table** and migration (`e7a1b2c3d4f5`), modelled on `references`: owner plus `read_visibility` / `write_visibility`, same authz path.
- **Gateway**: a new **Reference types** tab under Resources — create, edit, delete, with the in-use guard surfaced as a server error rather than a client pre-check.
- **Agent protocol**: `list_reference_types` returns the types the calling agent's owner can read. Each entry now also carries `value_hint` and `origin` (`builtin` or `user`).
- **Server-driven form metadata**: the gateway's client-side `VALUE_FORMS` registry is removed. A type's `value_hint` is what renders under the URL box, so a new type needs no frontend change to be usable.
- **Rooms YAML**: reference types are validated at provisioning time instead of failing later.

Slugs are a **global namespace**, not per-user. A room's reference set spans owners, so per-user slugs would collapse into a flat set at the point of use and become unresolvable.


amaudruz (changes requested):
Nice! Just one nit and something else I realize we could add.

Nit: 

Image

Placement of the delete and edit buttons here. Not sure if this also applies to the other tabs in this page tho

For the other thing we can do that I realised just now is missing: provide a new tool for agents to list all the existing external references ( with optional filter on name pattern, type, owner, etc.)  so that they can attach to rooms.

For now there is only one tool that allows them to list references attached to the current room they are in.

amaudruz (approved):
Actually nevermind the additional thing, let's add it to a separate task. Already pre approving but could be nice to fix the nit

raullopez-sandbox:
Nit fixed in `f5937180` ✅

PR #336, opened 2026-09-02
fix(sidecar): reap dead tmux targets

### LINKED ISSUE #321: sidecar retries injection into a dead tmux target indefinitely (17,425 times observed) and never reaps it — agent unreachable while showing live
## Summary

When a session's tmux target dies, the sidecar retries injection into the dead target indefinitely and never reaps it. Observed **17,425 consecutive failed injections** into a dead tmux target recorded in `state.json`.

The agent is unreachable throughout, while the room shows it **live** — the sidecar's own heartbeat keeps presence green, because presence reflects the sidecar rather than a session.

Related to #320 (a different silent-delivery failure), but independent and arguably more damaging.

## Behaviour observed

- A session's tmux target is killed (in our case deliberately, to test restart survivability).
- The sidecar restores that same session id from `state.json` and begins injecting into the now-dead target.
- Every injection fails. Nothing reaps the stale entry, and no new session is auto-started.
- No error surfaces in the room. The agent shows `live`, `last_seen_at` stays current, and `list_participants` reports it as present.
- Addressed messages receive no reply and no 👀 reaction.

## Why it is hard to diagnose

Every signal a user can see says the agent is healthy. The distinguishing probe is the **👀 reaction**: it appears only when a message reaches a live session, so its absence separates "sidecar up" from "session up". Room presence cannot make that distinction.

## Impact

Recovering requires resetting the remote agent (Switch Console). Notably, patching the connector config and relaunching does **not** help — the sidecar reloads the same dead session from `state.json` and resumes the loop, so the obvious recovery path appears to work and does not.

## Suggested fix

1. **Reap a dead tmux target.** Detect that the target no longer exists and clear the session from `state.json` rather than retrying against it.
2. **Bound the retries.** An unbounded loop with no backoff and no ceiling cannot be right; after N failures the sidecar should give up on that session.
3. **Fail loud.** Log a warning on the first failed injection naming the target, and surface the state so a user can see the agent is not merely idle. Per the repo's "Fail Loud, Never Fake" guidance, silently degrading to look fine is the outcome to avoid.
4. **Auto-start after reaping**, so a dead session self-heals instead of requiring a Console reset.

mikemikimike:
Fixes #321

The sidecar liveness poll now reaps runtime sessions as soon as their tmux target disappears. Reaping uses the existing stopSession lifecycle, which stops the RoomConnection no-target retry loop, clears startup-watch state, removes the durable registry entry, and releases the spawner and notification-watcher room guards so a later notification can auto-start a replacement.

Added regression coverage for dead-pane cleanup and for preserving unrelated live sessions.

Validation:
- pnpm run build:sidecar
- pnpm vitest run --project node src/sidecar/sidecar-runtime.test.ts src/main/core/switch-rooms/room-connection.test.ts (86 passed)
- pnpm exec tsgo --noEmit
- changed-file oxfmt and oxlint
- Full node/main-db run is environment-blocked by incomplete Electron/better-sqlite3 installation and existing Windows/POSIX assumptions (2444 passed, 53 failed, 56 suites blocked).

### DIFF
--- a/console/apps/switch-console-desktop/src/sidecar/index.ts
+++ b/console/apps/switch-console-desktop/src/sidecar/index.ts
@@ -413,6 +413,20 @@ async function main(): Promise<void> {
       })
     );
     for (const t of [...liveTargets]) if (!targets.has(t)) liveTargets.delete(t);
+
+    // A dead pane is terminal for the runtime connection. Reap it now rather
+    // than leaving RoomConnection's no-target retry loop and the durable
+    // session registry alive indefinitely. The watcher guard is released so a
+    // later notification can auto-start a replacement for the room.
+    for (const { sessionId, roomId } of runtime.reapDeadSessions()) {
+      spawner?.drop(sessionId);
+      if (roomId) watcher?.clearRoom(roomId);
+      log.error('sidecar: reaped session after its tmux target died', {
+        event: 'switch_session_reaped_dead_target',
+        sessionId,
+        roomId,
+      });
+    }
   };
   const refresh = async (): Promise<void> => {
     await Promise.all([refreshLiveness(), refreshWatchEnabled(), refreshLaunchSpec()]);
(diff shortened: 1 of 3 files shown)


mikemikimike:

amaudruz (approved):
Thanks for this ! @mikemikimike

PR #337, opened 2026-09-02
fix(console): give the bare destructive colour utilities a token

raullopez-sandbox:
## The problem

The theme does not define a `destructive` colour. The `@theme inline` block in
`src/renderer/index.css` exports `--color-background-destructive`,
`--color-foreground-destructive` and `--color-border-destructive`. It does not
export `--color-destructive`.

Tailwind writes a rule only for a class it can resolve. So `text-destructive`
resolves to nothing, and Tailwind writes no rule for it. The element keeps the
colour it inherits.

About 120 places in the renderer use these classes:

| Class | Uses | What happens now |
| --- | --- | --- |
| `text-destructive` | 66 | The text is the normal foreground colour. |
| `border-destructive`, and the `/30 /40 /50` variants | 19 | The border is the normal border colour. |
| `bg-destructive/5 /10 /20 /30` | 12 | There is no background. |
| `ring-destructive/20 /40` | 24 | `ring-3` still draws a ring, in `currentColor`. |

The last row is the worst one. An invalid input does not lose its ring. It gets
a black ring in the light theme and a white ring in the dark theme.

I found this while I worked on #325. The error message in the "New agent" modal
looked grey. Louis asked for red text. The class was already there.

## The change

Add one line to the theme:

```css
--color-destructive: var(--foreground-destructive);
```

`--foreground-destructive` is `--red-11` in the light theme and `--red-9` in the
dark theme. These are the values the sibling tokens already use for red text.

I chose this over a rename of the 120 classes. Two of the four families have no
other token to move to: `bg-destructive/10` and `ring-destructive/20` need a
colour with an alpha value, and the theme has no `background-destructive` shade
for that.

The second commit is formatting only. `oxfmt` sorts Tailwind classes, and it
sorts only the classes the theme defines. The new token makes `text-destructive`
a known colour utility, so it now sorts after `text-xs`. Every changed line in
that commit names a `destructive` class.

## What it looks like

I rendered a gallery of the 12 patterns that use these classes. Left is `main`,
right is this branch.

### Light theme

| Before | After |
| --- | --- |
|  |  |

### Dark theme

| Before | After |
| --- | --- |
|  |  |

## One thing to look at

Row 5b is `Button variant="destructive"`. In the dark theme the button now takes
`dark:bg-destructive/20`, because that class beat `bg-background-destructive`
before and did nothing. The fill moves from `--red-3` to a red tint. I think it
is an improvement, but it is a change of colour on a button that was already
visible. Please tell me if you disagree.

## How I checked it

I rendered the gallery in Chromium with the real Tailwind build, in both themes,
before and after. The screenshots above are that gallery.

I also read the computed style of a `text-destructive` span before the change.
It was `color(display-p3 0.933 0.933 0.933)` — the same as a plain span next to
it.


amaudruz (approved):
noice

PR #346, opened 2026-09-02
feat(bridges): render agents under their display name on Mattermost

raullopez-sandbox:
Stacked on #338. Two commits: the bridge change, then the deploy default.

## Why Mattermost needed its own PR

Every other platform lets a single bot post under a different name per
message. Mattermost doesn't — it gives **each agent its own bot account**, and
that account's username *is* the routing handle: the mention, the member-list
key, and how Switch re-adopts the bot after a restart. Usernames are also
3-22 chars, lowercase, no spaces, so "Switch Dev" cannot be one.

So the label goes on the bot's `display_name` and the username stays the
identifier, untouched.

Two things fall out of that:

- **Adoption was already broken.** When Switch re-adopts an existing bot it
  re-applies the icon but never touched the display name. Renaming an agent
  left the old label on the account forever. The adopt path now patches it
  when stale.
- **The name is a name field, not message text.** It takes the label
  verbatim, *not* the zero-width-defused body form. Sending the escaped form
  would bake invisible characters permanently into a stored account name —
  invisible in a diff, and there for good.

## The deploy commit

Mattermost renders a bot under its username unless `TeammateNameDisplay` says
otherwise, and the default says username. Without the second commit, Switch
writes "Switch Dev", the write succeeds, nothing errors, and every human still
reads `switchdev` — the silent-degradation case this repo explicitly forbids.

The bridge warns when it detects that (once, and only when a label would
actually be hidden). The deploy commit makes it *work* wherever Switch owns
the server: local compose, standalone compose, Helm chart, plus the pinned
copy Switch Console bundles — re-synced with the repo's own
`sync-standalone-compose.mjs`, whose `--check` mode gates CI and would have
failed the first three on their own.

**Reviewer note, deliberate trade-off:** `TeammateNameDisplay` is server-wide,
so human members of these deployments render under their full names too. And
because Mattermost freezes any setting supplied via env var, this becomes
fixed in the System Console rather than admin-editable. Both are consequences
worth accepting for the feature to do anything at all, but they are product
choices, not implementation details.

## Verification

Core: 2201 pass, ruff and mypy clean, compose drift check green.

I ran a controlled experiment against a real Mattermost 11.9.0 rather than
trusting the docs — same server, same bot, same post, only the setting
changed:

| `TeammateNameDisplay` | post header |
|---|---|
| `full_name` (this PR) | **Switch Dev** |
| `username` (stock) | **switchdev** |

The stored account record was identical in both — `username: "switchdev"`,
`first_name: "Switch Dev"`. That contrast is the entire argument for the
deploy commit.

I also mutation-tested the bridge commit. The one the suite initially missed:
swapping the verbatim label for the escaped body form passed all 1086 tests,
so I added `test_a_mattermost_bot_display_name_is_not_escaped_for_message_text`
to close it.

### DIFF
--- a/console/apps/switch-console-desktop/src/main/core/managed-switch-server/resources/standalone-docker-compose.pinned.yml
+++ b/console/apps/switch-console-desktop/src/main/core/managed-switch-server/resources/standalone-docker-compose.pinned.yml
@@ -139,6 +139,7 @@ services:
       MM_SERVICESETTINGS_ENABLEBOTACCOUNTCREATION: "true"
       MM_SERVICESETTINGS_ENABLEUSERACCESSTOKENS: "true"
       MM_TEAMSETTINGS_ENABLEOPENSERVER: "true"
+      MM_TEAMSETTINGS_TEAMMATENAMEDISPLAY: "full_name"
       # Neither plugin is usable in a bundled Switch deployment: Copilot has no
       # configured provider and throws a 404 looking up its bot on every page
       # load, and Calls ships a WebRTC service nothing here can reach.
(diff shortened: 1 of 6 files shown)


abeldantas (approved):
LGTM

PR #355, opened 2026-09-03
gitignore: add .claude/worktrees/

### LINKED ISSUE #332: .gitignore does not cover .claude/worktrees/, so agent worktrees are commitable
## What happens

`.gitignore` anticipates two things under `.claude/`:

```
.claude/settings.local.json
.claude/agents/
```

It does not cover `.claude/worktrees/`. When an agent is launched with worktree
isolation, the harness creates a full nested checkout at
`.claude/worktrees/agent-<id>/` **inside the repo directory**. `git status` then
reports `?? .claude/`, and a `git add -A` sweeps it toward the index.

Reproduce:

```
git check-ignore -q .claude/settings.local.json && echo IGNORED   # IGNORED
git check-ignore -q .claude/agents            && echo IGNORED   # IGNORED
git check-ignore -q .claude/worktrees         || echo "NOT IGNORED"  # NOT IGNORED
```

## Why it is worth a line of config

The other uncommitted directory in this repo, `internal/`, is untracked by
deliberate choice and a human put it there, so the people working in the tree
know to avoid it. `.claude/worktrees/` is different: **nothing a human did puts
it there.** It appears when a tool decides to isolate an agent, at a path the
ignore file was written before anyone anticipated. Someone who has been careful
about `internal/` for months gets a new commitable directory without being told.

The failure is also quiet in the worst way — a nested checkout added to the index
does not look obviously wrong in `git status`, and the damage shows up in a PR
diff rather than at the moment of the mistake.

## Suggested fix

One line:

```
.claude/worktrees/
```

Or, more durably, ignore the whole directory and un-ignore the parts that should
be shared:

```
.claude/*
!.claude/settings.json
```

so that the default for anything new a tool drops in `.claude/` is "ignored"
rather than "commitable". That inverts the failure mode: a new tool-created path
is silently safe, and sharing something is the deliberate act.

## Note

Filed rather than fixed inline because it is unrelated to any feature branch in
flight and is a repo-wide policy choice. Found when a worktree-isolated agent
made `.claude/` appear in a previously clean `git status`.

---

## Addendum — two things a reviewer will check, verified

**Nothing under `.claude/` is tracked.** `git ls-files '.claude/'` returns zero
files. This cuts both ways and the issue should be honest about it:

- It makes the inverted form **strictly safe to land**. There is no tracked file
  for `.claude/*` to newly ignore, so no file changes status and no history is
  disturbed.
- But it also means `!.claude/settings.json` un-ignores a path **that has never
  existed in the index**. The exception line matches nothing today. A reviewer
  who checks will find that and may reasonably read it as evidence the rule is
  wrong.

It is not wrong — it is a statement about what *should* be shareable if a shared
settings file is ever added, which is the point of a default-deny rule. But it
is a claim about the future, not a fix for something currently broken, and it
should be labelled that way rather than discovered. If that is unwelcome
speculation, `.claude/worktrees/` alone is the conservative fix and closes the
actual hole.

**`.claude-plugin/` is a different directory and is not affected.**
`.claude-plugin/marketplace.json` **is** tracked, and it is the connector
marketplace manifest — the file deciding whether the Claude Code and Codex
plugins are installable at all. `.claude/*` does not glob into
`.claude-plugin/` (different directory name, not a prefix match), so the
inverted form leaves it alone. Worth stating explicitly in the issue because
"ignore the agent-tooling directory" is precisely the change that could quietly
eat a manifest, and had the directory been named `.claude/plugin/` this would
have been a considerably worse idea.

## Relationship to #328 (corrected)

An earlier version of this issue claimed a "`git status` is clean after a build"
CI gate would have caught this. **That is wrong, and it is worth correcting
explicitly rather than quietly deleting.**

This failure happens on a developer machine, at development time, when the agent
harness creates `.claude/worktrees/`. **CI never runs Claude Code**, so that
directory never exists in a CI checkout. A "working tree clean after build" gate
therefore passes in exactly the world where this bug is live — it is a no-op
here. (Such a gate is worth having for unignored *build artifacts*. That is a
different problem.)

The gate that does work asserts on the **index**, not the worktree:

```
git ls-files -- '.claude/' | grep . && {
  echo "tracked files under .claude/ — see #332"; exit 1; }
```

That catches the actual bad outcome — someone committed a harness-created
directory — rather than the precondition. Verified against this repo: it returns
nothing today, and the `.claude/` pathspec does **not** match
`.claude-plugin/marketplace.json`, so it cannot false-positive on the connector
manifest.

So the honest grouping with #328 is narrower than first stated:

- **Shared:** both are invisible-absence failures. The thing to notice is
  something *not happening* — a migration chain never applied, an ignore rule
  never covering a path — and code review is structurally poor at noticing
  absences.
- **Not shared:** only #328 is fixed by a CI gate over *behaviour* (apply the
  chain from empty; a broken migration then fails at the commit that introduces
  it). #332 is fixed by the ignore rule itself, with the index assertion above
  as a backstop that stops it recurring.

They are the same shape of problem and different shapes of fix. Grouping them
for triage is reasonable; claiming one gate covers both is not.

Dahhrk:
## Summary

Add `.claude/worktrees/` to `.gitignore` so that agent worktree directories are not committable.

The `.gitignore` already covers `.claude/settings.local.json` and `.claude/agents/` but not `.claude/worktrees/`, which means agent worktree directories show up as untracked files and can be accidentally committed.

## Change

One line added to `.gitignore`:

```
.claude/worktrees/
```

## Verification

- `git check-ignore -q .claude/worktrees/` → exit 0 (ignored)
- `git check-ignore -q .claude/worktrees/somefile` → exit 0 (files inside ignored)
- `git ls-files .claude` → empty (no tracked files under `.claude/`)

Fixes #332

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -231,6 +231,7 @@ __marimo__/
 related-projects
 .claude/settings.local.json
 .claude/agents/
+.claude/worktrees/
 
 # Switch Console working-tree artifacts. Build output and scratch downloads that
 # land at the repo root; ignored so they cannot be committed by accident.


abeldantas (approved):
One line, does what it says.

abeldantas:
@Dahhrk if you want to contribute you should say:

Verbatim without a period as comment to this pr so we can go ahead and merge this.
(After reading and accepting the CLA ofc)

PR #365, opened 2026-09-03
feat(db): record every message into Postgres alongside the send, and verify it (CHOO-1436)

amaudruz:
**Stacked on #363** (`worktree-transport-port`). Review that first; the diff here is only the two commits on top.

## Why

Step 3 of the Tuwunel-replacement plan: give Switch a Postgres record of every message, written alongside the send. Nothing reads it yet — the homeserver is still the source of truth for history — so this is the low-risk half of the migration. It is also worth having on its own merits: queryable history, full-text search, retention policies and per-tenant deletion are all things Switch cannot do today because messages only exist inside RocksDB.

## Where the write goes, and why there

The narrowest point that sees 100% of outbound is `MatrixTransport._room_send`, but it has no session factory and no Switch identities — only a Matrix room id and an mxid. The narrowest point that *has* identities is `ClientBase`, which until now missed ~20 call sites that reached past it to `transport.send_event` directly.

So `ClientBase` gained the `send_event` wrapper it was always missing, those call sites now go through it, and the three wrappers are instrumented. That is also exactly the seam step 5 needs.

**Outbound-only capture is complete, and this is a finding rather than an assumption.** `room_send` appears nowhere outside `transport/matrix.py`, and `matrix_admin.py` only provisions rooms and users — it never sends. Every participant in a Switch room is a Switch-owned client, so every message is somebody's outbound, including a bridged human's (which enters through their puppet). Capturing inbound as well would write N-1 duplicate rows per message in an N-client room.

## Schema

`messages` and `message_attachments`, following the existing conventions (Text uuid PKs, timestamptz, JSONB, explicit FK cascades).

`content` holds the event body as sent; the columns beside it are denormalised out of it for querying. For a `com.switch.*` event they are mostly null and `content` carries everything, which is why it is `NOT NULL`.

## Three decisions worth arguing about

**`SendResult` now carries `event_type` and `content`.** The transport assembles the wire content, so the caller never sees it. Recording it meant either reporting it back or rebuilding it at the client — and a rebuild drifts from the transport the first time the wire format changes, silently. The transport reports what it sent.

**`seq bigint`, not in the plan as written.** Step 4 pages on this. `sent_at` ties between messages sent in the same transaction, and a cursor with ties can skip or repeat rows across pages. Adding an ordering column later means backfilling one.

**`delivery_cursors` deliberately not added.** The plan lists it, but nothing in steps 3 or 4 writes or reads it and its shape is decided by the step 5 dispatcher design. A table nobody writes is a guess made a month early.

## The failure mode, stated rather than hidden

The Matrix send happens first and is unchanged. The Postgres write follows and **cannot fail it** — a database problem must not make messaging less reliable than it is today, which is the whole selling point of this step being low risk.

The consequence is real and worth naming: a write that fails after a successful send leaves a row missing for a message that was really delivered. It logs at `error` rather than being swallowed, and step 4's "diff Postgres against Matrix on live rooms before flipping" is what catches the divergence.

`ClientBase.send_event` raises where `send_message` returns `None`. A chat message the homeserver rejects is a message nobody reads; a `com.switch.*` event that never lands is protocol state a caller is about to assume exists.

## Pre-existing drift this surfaced — not caused here

Nothing in the suite ever ran the migrations: the test schema is built from the models with `create_all`, which `test_migration_chain.py` says outright in its own docstring. So a model could drift from its migration and every test still passed.

`test_migration_parity.py` closes that — it replays the chain against real Postgres and diffs it against `Base.metadata`. It immediately found four indexes that migrations create but the models never declare:

- `ix_agent_sessions_agent_room`
- `ix_agent_sessions_transport_session_id`
- `ix_agents_parent_agent_id`
- `ix_external_user_claims_user_id`

Autogenerate would emit `DROP INDEX` for all four today. They are allowlisted **by name**, so a fifth divergence still fails the test. The fix is four `Index(...)` declarations and belongs in its own change, not buried in this one.

The new tables are in exact parity with their migration.

## Testing

- **2290 passed.** ruff and mypy clean.
- `test_message_store.py` — 16 tests against real Postgres, including the FK behaviours (`ON DELETE CASCADE` for a room's messages and a message's attachments, `SET NULL` for a deleted client) and `seq` paging with no gap or repeat across pages.
- `test_message_recorder.py` — what gets written for each content shape the transport really produces, and the two failure paths: an unknown room warns, a database failure logs at error and never reaches the caller.
- `test_send_recording.py` — each send route is checked separately, because instrumenting one says nothing about the others; and a send the transport rejected records nothing.
- The caching test was verified to fail when the cache is removed. A test that passes either way is worse than none.

## Not verified

**The integration suite against the real Matrix stack has not been run** — same caveat as #363, and it carries over.


amaudruz:
Second half pushed: `d89b8307`.

Adds the reconciliation check the PR body said was missing — `just reconcile-messages`, which walks each room's history on the bus, diffs it against the recorded rows, and exits non-zero on drift. Logic lives in `switch_core/messages/reconcile.py` so it is testable; the script is thin wiring following the `scripts/gen_artifacts.py --check` convention.

Three decisions in it worth reviewing, all of the same kind — refusing to let the check report a falsely clean answer:

- **The window is bounded to where recording was active**, starting at the oldest recorded row in each room. Rows begin when the recorder was deployed, so a room's whole back-catalogue is legitimately absent; reporting it would bury the one row that actually went astray.
- **The history walk is unbounded**, unlike `read_context`, which stops at `HISTORY_MAX_PAGES`. A reconciliation that gave up early would report the events it never looked at as agreeing.
- **A room the reader is not in is a failure, not a skip**, and a room with nothing recorded reports "nothing to compare" rather than "clean".

Fields are compared rather than whole content dicts: the bus can hand back an event annotated with fields it added itself, and flagging those would train everyone to ignore the report.

2301 passing, ruff and mypy clean.

The caveat that remains is unchanged and is now the only one: **neither this nor #363 has been run against the real Matrix stack.** This tool is the thing that would answer it, and it needs a deployment to run against.

PR #383, opened 2026-09-06
fix(mcp): preserve operation return docs

### LINKED ISSUE #348: MCP door drops every operation's Returns block, so agents on that path see no documented return shape
## What

`bridges/agent/operations/registry.py:73` registers each operation with its **full** docstring:

```python
description=(fn.__doc__ or "").strip(),
```

The HTTP `/ops` surface serves that verbatim, so a client there gets the `Args:` and `Returns:` blocks. The MCP door (`bridges/agent/mcp/server.py`) hands the same functions to FastMCP, which parses the docstring, lifts `Args:` into the input schema, and **discards everything from the first section marker onward**. The `Returns:` block never reaches an MCP client.

## Observed

Dumping the live surface for the same two operations:

`await mcp.list_tools()` → description for `list_agents`:

```
List the agents registered on this Switch instance.

Unlike `list_participants` (which is scoped to the connected room), this
lists every agent on the instance. Use the optional filters to narrow the
result; they are ANDed together, and omitting one ignores it.
```

`all_operations()['list_agents'].description` → the same operation, HTTP door:

```
List the agents registered on this Switch instance.
...
Args:
    name_contains: Case-insensitive substring to match against agent names.
    ...
Returns:
    A list of agent summaries (sorted by name), each
    {id, name, description, icon_url, display_name, connector_type, ...}
    `icon_url` is null when the agent has no icon set. `display_name` is
    null when the agent has no display name set; fall back to `name`.
```

This is not specific to `list_agents` — it applies to every operation with a `Returns:` block.

## Why it matters

The `Returns:` block is where the operation surface documents its response shape, and for several operations it carries the only statement of a rule an agent has to follow — which fields can be null and what to do about it, which identifier addresses and which is decorative, which id to pass on to the next call. An agent on the MCP path is told none of it and has to infer the shape from a sample response, which works right up until it hits the case the prose was warning about.

The impact is bounded by the fact that the MCP door is the legacy path — `mcp/server.py`'s own docstring says the supported direction is the local runtime beside the agent — so the fix may well be "nothing, it is on the way out". Filing it because the asymmetry is silent: the two doors serve visibly different documentation from one source and nothing says so.

## Found via

Verifying that a docstring change in #347 actually reached agents. It reaches the HTTP surface; it does not reach MCP.

bytehello:
## Summary

- pass each registry operation’s full description explicitly to FastMCP
- preserve `Returns:` contracts on the MCP surface without changing generated input schemas
- add focused coverage comparing the MCP description with the source operation contract

## Testing

- `uv run --project core ruff format --check .`
- `uv run --project core ruff check .`
- `uv run --project core mypy --config-file core/pyproject.toml core/switch_core/ connectors/`
- `uv run --project core pytest -c core/pyproject.toml core/tests/switch_core/bridges/agent/test_mcp_tool_surface.py core/tests/switch_core/bridges/agent/api/test_operations.py -q` (20 passed)

The full non-integration test command reached 2,145 passing tests; 504 database-backed tests could not initialize because no Docker daemon is available in this environment.

Closes #348

### DIFF
--- a/core/switch_core/bridges/agent/mcp/server.py
+++ b/core/switch_core/bridges/agent/mcp/server.py
@@ -84,7 +84,7 @@ mcp.add_middleware(CallContextMiddleware())
 # Every operation, registered as a tool. This loop is the only thing that makes
 # an operation an MCP tool, so the two surfaces cannot diverge.
 for _op in all_operations().values():
-    mcp.tool(_op.fn)
+    mcp.tool(_op.fn, description=_op.description)
 
 logger.debug("Registered %d operations as MCP tools", len(all_operations()))
 
--- a/core/tests/switch_core/bridges/agent/test_mcp_tool_surface.py
+++ b/core/tests/switch_core/bridges/agent/test_mcp_tool_surface.py
@@ -13,6 +13,7 @@ from pathlib import Path
 import pytest
 
 from switch_core.bridges.agent.mcp.server import mcp
+from switch_core.bridges.agent.operations import all_operations
 
 TASK_PROTOCOL_TOOLS = {
     "delegate_task",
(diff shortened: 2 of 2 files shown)


bytehello:

bytehello:
recheck

amaudruz (commented):
thanks!

PR #392, opened 2026-09-08
Multi-tenancy Phase 0: tenant id on every log line, and the RDS groundwork (CHOO-2622)

petr-sandbox:
Phase 0 of multi-tenancy: make the logs able to say *which* tenant, and get Postgres ready to move to RDS. No behaviour changes for anyone using Switch today.

## Logging

Switch had no logging setup at all. Logging worked only because Alembic's `fileConfig` ran at startup and configured the root logger as a side effect — which is why log level came from two ad-hoc `os.environ` reads and nothing else was configurable.

So rather than plumb a tenant id through ~470 call sites, this adds the layer that was missing:

- A context lives in a `contextvar`, and a filter on the root handler stamps it onto every record. **No log call site changes**, and records from libraries (uvicorn, SQLAlchemy) carry the fields too.
- `LOG_FORMAT=json` emits Datadog's standard attributes — `status`, `message`, `service`, `env`, `version`, `logger.name`, `error.kind/message/stack` — plus `tenant_id`, `request_id`, `agent_id`, `user_id`. `text` stays readable for a terminal.
- `TENANT_ID` is a deployment-wide setting for now, as agreed; the mechanism already prefers a bound value, so Phase 1 only has to change where the tenant comes from, not how it reaches the logs.
- Request ids come from a pure-ASGI middleware (`BaseHTTPMiddleware` runs the app in a separate task, where a contextvar set in the middleware is invisible). An inbound `x-request-id` is honoured but sanitised — a header cannot inject a newline into the log stream. Agent id is bound at bearer auth, user id at gateway auth, and collaboration bridge events get a synthetic id so an inbound Slack message can be followed across the handlers.
- `main()` now configures logging deliberately, and Alembic's `env.py` defers to it in-process while keeping its own setup for the CLI.

## RDS

- `DB_SSL_ROOT_CERT` — a verifying SSL mode previously had no way to be told which CA to trust, so `verify-ca`/`verify-full` could not validate a managed database's certificate at all. The connection is now built with an `SSLContext`. A missing file, or a bundle set alongside a non-verifying mode, fails at startup.
- `docs/old/rds-migration.md` — the proposal and cutover runbook. **Nothing is provisioned and nothing was applied.**

Two things in the runbook are worth reading before anyone provisions:

- **Do not put a transaction pooler in front of it.** Message delivery is a `pg_notify` trigger plus a held `LISTEN`. RDS Proxy pins a connection permanently on LISTEN and offers no pinning filter for PostgreSQL; pgbouncer in transaction mode never supports it, and fails *quietly* — `NOTIFY` still works, so messages are written and announced to nobody.
- **`verify-full` requires the RDS endpoint hostname**, not a friendlier CNAME. Wanting a stable internal name means accepting `verify-ca`, which is a real downgrade.

The one gap left open deliberately: the Helm chart has no `DB_SSL_ROOT_CERT` and no way to mount the bundle, so a deployment cannot reach `verify-full` until that lands. Chart work was agreed to come later.

## Deployment

`switchCore.logging` carries the format, root level, tenant, service and environment names into switch-core and the migration Job — without it the logging above is unreachable from a deployment, which is the point of the ticket. Defaults are today's behaviour.

`postgresql.caBundle` carries the CA, inline or as an existing ConfigMap, and mounts it into everything that connects (switch-core, the migration Job, Mattermost). The chart refuses a configuration that cannot work — a verifying `sslMode` with no bundle, a bundle under a mode that ignores it, a log format the app would reject — at render time rather than as a crash loop.

## Rehearsed, not just reasoned about

The cutover was walked through locally against two throwaway Postgres 16 containers, one TLS-enabled and owned by a non-superuser like RDS's master. The restore carried every index, foreign key and the notify trigger, and a `bytea` blob byte-identical; `verify-full` connected against the supplied CA and refused an unrelated one; and an insert over `verify-full` reached a live `LISTEN` subscriber — trigger, listener and TLS at once, on a restored database. Three corrections came out of it and are in the runbook: `--exit-on-error` is needed or a non-superuser restore reports its failures and carries on, the hostname trap is real, and the schema has no sequences at all, so that class of cutover bug cannot happen here.

## Testing

21 new tests (logging config, context propagation into spawned tasks, request-id handling including a hostile header). Full suite green: 2579 passed. mypy clean.

Nothing here turns on until someone sets a value: the CA bundle is inert unless configured, and the logging defaults keep today's behaviour apart from the extra field on each line.


petr-sandbox:

amaudruz (approved):

PR #394, opened 2026-09-08
Rich session activity in the messaging apps, Slack first (CHOO-2621)

simonflk-sb:
## Purpose

Someone watching an agent work from Slack cannot tell what it is doing, and
cannot answer it when it asks for something — the only signal today is a
plain-text "working…" indicator. This builds the layer between the session
interaction contract and the messaging platforms that turns a session's
requests and turn activity into a real, answerable presence in a channel,
Slack first.

## Key Changes

- **A platform-neutral session projection.** The contract is mirrored into
  Python (a byte-identical lift of `session-v1` from the upstream SDK branch)
  and held as a per-session replica: coalescing, dedup by `(itemId, revision)`
  / `(requestId, revision)`, and deciding channel root versus a secondary
  surface.
- **Request cards on Slack.** An approval or a set of questions renders as
  buttons plus a text fallback that is answerable by typing (`R1 2`), backed
  by a handle reserved per request so an answer always resolves to the right
  card and revision.
- **Turn activity as one message per turn**, edited in place rather than
  reposted, ending on the turn's own state — streamed into a thread as a
  `plan` block where Slack allows it, posted and rewritten at the channel
  root otherwise.
- **A verified inbound path.** A press or a typed answer becomes a `Command`
  from a verified Switch identity, never from the payload; a refusal (stale
  card, unresolvable actor) is told back to the person, threaded under the
  card, on every collaboration platform — not just Slack.
- **`post_rich` / `update_rich` on the adapter port**, so which renderer draws
  a turn or a card is each adapter's own choice. The base falls back to a
  short neutral turn summary for a platform with no renderer of its own yet;
  a request card has no neutral form yet and is stubbed pending the first
  off-Slack renderer.
- **`!session-demo`**, behind a feature flag, replays a recorded fixture
  through the whole path into a real bridged channel — the way this has been
  verified against actual Slack behaviour with no backend in place yet.

## Design Decisions

- The existing runtime-state indicator is untouched and stays as the fallback
  for connectors that cannot speak the contract.
- Room interaction (posting, inviting, attachments) stays on the MCP server,
  out of the session replica — session activity and room interaction are
  different things.
- Out of scope here: the server half (routes, storage, leases, reservations —
  PR #411). This branch is driven from fixtures until that lands, so nothing
  yet consumes the `Command`s it builds.

## Testing

Full test suite passes locally (2,900+ tests, including the Postgres-backed
ones), with no change to Slack's rendered output guarded by the existing
Slack-specific suite passing unmodified throughout. Manually exercised
against a real Slack workspace via `!session-demo`.

**Not for merge yet.** Targets a temporary base branch so only this branch's
code is under review; lands once the upstream contract is accepted and the
server half (#411) is in place.


core/switch_core/bridges/collaboration/session/renderers/slack.py
```diff
@@ -0,0 +1,140 @@
...
+    content = request.content
+    if not isinstance(content, ApprovalContent):
+        raise ValueError(
+            f"Request {request.request_id} is not an approval: {content.kind}."
+        )
+    lines = [f"> Request {reference.handle}: {content.title}"]
+    if content.detail:
+        lines.append(f"> {content.detail}")
+    lines += [
+        f"{index}. {option.label}"
```
simonflk-sb: Do we need to escape these values?

simonflk-sb: > Sent by :robot: `@simonflk.remote-coder`  from Switch:

Yes for that line, no for the buttons. Good catch — it was a real hole.

`render_approval_text` becomes the `text` field of `chat.postMessage`, and **Slack parses `text` as mrkdwn whether or not blocks accompany it**. So an approval's title, detail or option label could write a `<!channel>` broadcast through the fallback that the card itself already refused. Fixed in 4ea6c9fd: every interpolated value in the text form is escaped, and only the `>` quote markers and the numbering are markup this function wrote.

**Button labels I have deliberately left alone.** They are `plain_text` block objects, which Slack does not parse — there is no markup to forge there — so an escaped `&` would reach the reader as a literal `&amp;`. They are also truncated at Slack's 75-character limit, which would cut an entity in half.

While in there I moved the three replacements to `slack/mrkdwn.py`. `escape_label_for_body` had the only other copy and now calls it, so there is one implementation of a security-relevant primitive rather than three.

Tests: `test_agent_text_cannot_forge_markup_in_the_text_fallback` forges into the title, the detail *and* both labels and asserts all four are neutralised; `test_a_button_label_is_left_as_the_author_wrote_it` pins the deliberate exception.

**One thing I found and did not fix.** A detail containing a backtick breaks out of the code span the card wraps it in — `` `pnpm test` `` becomes unbalanced. Slack documents no escape for its emphasis characters, which is the same limitation `escape_label_for_body` already calls out as cosmetic, and the markup that could actually be forged is closed. Say the word if you would rather the detail moved to a `rich_text_preformatted` block, which is not parsed at all.

simonflk-sb:

core/switch_core/bridges/collaboration/session/outbound.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class SessionRequestCards:
+    """One bridge's posted request cards, as the requests behind them change."""
+
+    def __init__(self, adapter: SlackAdapter) -> None:
+        self._adapter = adapter
+
+    async def refresh(self, post: SessionRequestPost, request: SnapshotRequest) -> None:
+        """Redraw the card for `request` where it was posted.
```
simonflk-sb: Editing in place is right while a request is being answered, but not when the host reopens one... the card's scrolled off and it flips back to live buttons where nobody's looking.

Can we settle the old card into a static state and post a fresh one instead? That also fixes the stale-version problem, since a new card gets a new row. Note the DB only allows one row per request today, so that constraint has to change.

core/switch_core/bridges/collaboration/session/inbound.py
```diff
@@ -0,0 +1,140 @@
...
+        actor_id = await self._identify(interaction)
+        if actor_id is None:
+            logger.warning(
+                "Ignoring an answer to request %s: no Switch identity for %s on "
+                "bridge %s. An answer is only ever attributed to a verified actor.",
+                post.request_id,
+                interaction.sender_id,
+                self._bridge_id,
+            )
+            return None
```
simonflk-sb: When a press can't land, the person who pressed needs to be told... ephemeral message, not just a log line. Doesn't apply yet since commands are dropped, but let's not lose it.

core/switch_core/bridges/collaboration/session/renderers/slack.py
```diff
@@ -0,0 +1,243 @@
...
+    request: SnapshotRequest, reference: RequestReference
+) -> SlackMessage:
+    """Render an approval as the card that stands for it right now.
+
+    Open, it offers a button per option and is also answerable in words. Once
+    an answer is in flight or the request has settled, the buttons go and the
+    card says what became of it instead.
+    """
+    content = _approval(request)
+    prompt = f"*{_HEADINGS[request.state]}*\n{escape_mrkdwn(content.title)}"
```
simonflk-sb: `prompt` is unbounded and Slack caps section text at 3000 characters, so a long title or detail fails the whole post. Needs truncating... and watch the order, cutting after escaping can split an entity in half.

simonflk-sb: > Sent by :robot: `@simonflk.remote-coder`  from Switch:

Fixed in 349aeb99. Every agent-supplied value now has a budget: title 1500, detail 1200, label 150, set so no combination of them can reach the 3000 the section allows however both are spent.

On the ordering: the cut is made on the **source** and the result escaped whole, never the reverse. `escape_mrkdwn` only ever lengthens, so the longest prefix that still fits is found by binary search on the source and then escaped, which cannot split an entity. The footer is the one place a plain truncate is right, because the whole line is escaped after assembly.

Tests, each failing before the change: a 9000-character title leaves the section at or under 3000; 4000 `&` characters (which escaping expands fivefold) fit and `re.sub(r"&(amp|lt|gt);", "", section).count("&") == 0`, so no half entity survives; the text fallback stays bounded with 25 long labels on top of a long title and detail; and a settled card's footer cannot be stretched by the chosen option's label.

core/switch_core/bridges/collaboration/session/inbound.py
simonflk-sb: > Sent by :robot: `@simonflk.remote-coder`  from Switch:

Agreed, and not implemented here. Nothing writes `session_request_posts` yet, so there is no real card to press and no reachable path to the refusal — an ephemeral now would be code no test can drive.

Recorded as **slice 8 — "Telling the actor when their answer cannot land"** in the PR description, ticked off with the rest, rather than a TODO in the source. It covers all four silent refusals in `command_for`, not just this one: unknown token, no Switch identity, superseded revision, and the session rejecting the command once the route exists.

core/switch_core/bridges/collaboration/session/contract.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def event_bytes(event: Any) -> int:
+    """The size the host measured.
+
+    Separators are pinned so this counts what `JSON.stringify` counts: Python's
+    default `json.dumps` pads every separator and would read a compliant event
+    as oversized.
+    """
+    return len(json.dumps(event, separators=(",", ":")).encode("utf-8"))
```
simonflk-sb: json.dumps defaults to ensure_ascii=True, so an emoji counts as 12 bytes here and 4 on the host

simonflk-sb: > Sent by :robot: `@simonflk.remote-coder`  from Switch:

Good catch, and worse than it looks: it is not only a mismatch, it is a mismatch in the rejecting direction. An event the host sent well inside 64 KiB gets refused on arrival here, and the less ASCII it contains the sooner. A message that is mostly emoji or CJK hits the cap at roughly a third of the size the host allows.

Fixed in 870990c9 — `ensure_ascii=False`, so what is counted is UTF-8 of the same string `JSON.stringify` produces, which is what `TextEncoder().encode(...).byteLength` counts in `validation.ts`.

The test has no TypeScript counterpart, because the divergence is in Python's defaults rather than in the contract: it asserts `event_bytes({"a": "😀"}) == len('{"a":"😀"}'.encode())` and that a notice of 10,000 emoji parses instead of raising `PAYLOAD_TOO_LARGE`. Both fail before the change.

simonflk-sb:
merged in #507

PR #397, opened 2026-09-08
fix(gateway): link a verified OIDC identity to its account by email

wojtyniakAQ:
## Summary

CHOO-2624 (multi-tenancy Phase 2) requires accounts to be keyed on verified
email, not on login method: someone who signs up with a password and later
signs in with an IdP sharing that verified email must land in the same
account, not a second one — otherwise a later change of login policy forces
a manual account merge.

Today `get_or_create_oidc_user` refuses that case with `OidcIdentityConflictError`
(409 at the callback), added in a previous security fix that closed a real
account-takeover bug (unverified/attacker-set email logging in as an existing
user, e.g. the seeded admin). This PR keeps that fix's protection but narrows
the refusal to exactly the case that's actually unsafe.

## What changed

- `get_or_create_oidc_user` now:
  1. Looks up the identity by the immutable `(iss, sub)` pair first, as before — a
     bound identity is never re-evaluated by email.
  2. If not found, and the *verified* email matches an existing account, links
     this identity to that account instead of refusing.
  3. If the email is **not** verified and it matches an existing account, it
     still refuses (`OidcIdentityConflictError` / 409) — an unverified email is
     attacker-controllable and must never pick an account, including when
     `GATEWAY_OIDC_REQUIRE_EMAIL_VERIFIED=false` lets an unverified login
     through at all (that flag only decides whether an unverified login is
     accepted for *provisioning or continuing to match* an identity; it never
     turns on linking a new identity into a pre-existing *different* account —
     see the caveat under "Fixes, round 2" below, though: a legacy sub-only
     identity is a separate path that bypasses this entirely).
  4. If the email is new, provisions a fresh user (`password_hash=None`,
     stored lower-cased) either way.

- Identity storage moves out of a single `oidc_iss`/`oidc_sub` pair in
  `users.metadata` into a new `oidc_identities` table: one row per linked
  `(iss, sub)`, unique on that pair. The old shape could only ever name one
  IdP identity per user; a user who links a second IdP (or a second account
  at the same IdP) now gets a second row instead of overwriting the first or
  being blocked.

- A migration (`04f27f37e474`) creates the table, backfills existing
  `oidc_iss`/`oidc_sub` pairs (including sub-only legacy rows) out of
  `users.metadata` into it, and adds a case-insensitive index on
  `users.email`.

## Fixes, round 1 (adversarial review before merge)

**A. Legacy sub-only rows were a lockout and a silent account fork
(regression from main).** The first version of the migration left rows whose
`metadata` had `oidc_sub` but no `oidc_iss` untouched, and nothing read that
key again. Two consequences: (1) with `GATEWAY_OIDC_REQUIRE_EMAIL_VERIFIED=false`
— which exists specifically for IdPs like Okta's org authorization server
that never emit `email_verified` for directory users — such an account would
now get a permanent 409 instead of the login main gave it; (2) if that
identity's claimed email had since changed, the sub-only match was gone and
the email match missed too, silently provisioning a brand-new `role="user"`
account and orphaning the original (possibly admin) one. Fixed by backfilling
sub-only rows into `oidc_identities` with a NULL issuer, and teaching
`get_by_oidc_identity` to match on `sub` alone when no issuer is stored,
backfilling the issuer once a login supplies one.

**B. Email matching was case-sensitive**, silently defeating the ticket.
Fixed with a `lower()` compare backed by a functional index (refined further
in round 2 — see below).

**C. A lost race returned a 500.** Two concurrent first logins for the same
`(iss, sub)`, or for the same brand-new email, raised an uncaught
`IntegrityError`. Fixed with a catch-and-retry (refined to use a savepoint in
round 2 — see below).

**D. Linking was silent.** `_link_identity` now logs (moved to after the
flush in round 2 — see below).

**E. The migration wasn't honest about its downgrade.** `downgrade()`'s
docstring now states the data loss directly (corrected further in round 2).

**F. Test gaps.** Added a real migration-chain replay over seeded data
(`test_migration_oidc_identities_backfill.py`), a DB-level unique-constraint
test, deterministic race tests, and restored a successor for the deleted
legacy-row test.

**G. Test comments narrated the change rather than the rule.** Reworded.

## Fixes, round 2 (adversarial re-review of round 1)

**1. NULL-issuer rows escaped the unique constraint entirely.** `UNIQUE(iss, sub)`
does not constrain a NULL issuer at all — Postgres treats every NULL as
distinct from every other NULL — so two different users could each hold a
legacy sub-only identity for the same subject, both insert cleanly, and
`get_by_oidc_identity`'s `.scalars().first()` (no `ORDER BY`) would resolve to
whichever one it saw first, backfilling the issuer onto it and leaving the
other permanently unreachable: one account silently annexed, one silently
orphaned, decided by row order. Fixed with:
  - A partial unique index, `ix_oidc_identities_sub_null_iss` — `UNIQUE(sub)
    WHERE iss IS NULL` — so at most one such row can exist per subject.
  - A migration pre-flight check that refuses to proceed (raises, does not
    silently pick a side) if duplicate legacy subjects already exist in the
    data, with a message naming the colliding subject(s) for a human to
    resolve by hand.
  - `get_by_oidc_identity`'s legacy lookup switched from `.scalars().first()`
    to `.scalar_one_or_none()`, since uniqueness is now guaranteed — a second
    match would mean the index failed to do its job and should raise loudly,
    not be silently resolved.
  - New tests: a DB-level test that the partial index rejects a second
    NULL-issuer row for the same subject, and a migration test that the
    pre-flight refuses when duplicate legacy subjects are seeded (verified
    manually too, against real Postgres, alongside the happy path and a
    direct-insert check of the partial index).

**2. Section H undersold the exposure and two docstrings overclaimed.** Fixed
by making H explicit that the legacy sub-only path bypasses email entirely
(see the updated H below), and rewording `config.py`'s docstring, the "only a
verified email may link" framing, and `models.py`'s `OidcIdentity` docstring
to state the legacy carve-out rather than imply an absolute. `models.py` also
now says explicitly what breaks the "sub is unique per issuer" assumption
(a second configured issuer) instead of asserting it as a permanent
invariant.

**3. `get_by_email` traded a loud failure for a silent choice.** Switching
from `scalar_one_or_none()` to `.scalars().first()` in round 1 fixed the
crash but introduced an undisclosed, non-deterministic pick between two
pre-existing case-duplicate rows (possible because `users.email` is still
case-sensitively unique and the new index is not unique — duplicates were
already possible before this PR and are not newly created by it, but the
*read* path previously would have crashed on them instead of quietly picking
one). Fixed: `get_by_email` now orders deterministically and logs a
`logger.error` disclosing the ambiguity when more than one match is found,
rather than resolving it silently. New user creation via
`get_or_create_oidc_user` also stores email lower-cased, so this method
doesn't create new case variants going forward (pre-existing rows of any
casing are untouched). New test:
`test_pre_existing_case_duplicate_emails_resolve_deterministically_and_log`
(the original version of this fix and its test both had a bug — see round 3,
item 1, below).

**4. The race retry rolled back the caller's whole transaction.** `_recover_from_race`'s
`session.rollback()` undid everything in the session, not just the failed
write — the same class of bug `room_store.py` documents from production
(`add_client`'s comment on the second write's `IntegrityError` rolling back
an otherwise-fine transaction). Switched to `session.begin_nested()`
savepoints around each risky write, matching the pattern already established
in `reference_type_store.py`. Latent today since the OIDC callback's session
holds nothing else, but this is a public store method.

**5. Three smaller fixes:**
  - `OidcIdentityConflictError` no longer doubles as both "genuine security
    rejection" and "transient race, retry"; the latter is now a distinct
    `OidcIdentityRaceError`, mapped to `503` (with `Retry-After`) at the
    callback instead of `409`, so an operator's logs/dashboards can tell a
    retry storm apart from actual rejected takeover attempts.
  - `_link_identity` now logs only after `session.flush()` succeeds, so a
    losing racer never logs a link that didn't happen.
  - The `downgrade()` docstring's claim that data loss is decided by
    "whichever row this UPDATE visits last" was wrong — Postgres does not
    define an order for which source row an ambiguous `UPDATE ... FROM`
    applies. Reworded to say the choice is unspecified, not sequential.

## Fixes, round 3 (adversarial re-review of round 2)

**1. The "deterministic tiebreak" test was flaky, and the property it
claimed was not actually implemented.** Round 2's fix ordered
`get_by_email`'s duplicate-resolution by `(created_at, id)`. Postgres's
`now()` is fixed for the whole transaction, so two case-duplicate rows
created in the same transaction — exactly what the test did — tie on
`created_at` and the real tiebreak was `id`, a random UUID: the outcome
varied between runs (observed 5/12 passing before this fix), and on a
losing run the pick was the opposite of what the docstring and test claimed
("oldest wins"). Fixed by dropping `created_at` from the ordering entirely
and ordering by `id` alone, with the docstring and log message now stating
plainly that this is an arbitrary-but-stable tiebreak, not a claim that the
lower id is the "real" or original account. The test no longer asserts an
age relationship; it asserts the pick is stable across repeated calls (and
no longer assigns either duplicate a role that implied one should win). Run
12 times after the fix: 12/12 passing.

**2. The migration pre-flight covered only half the collision space.** The
duplicate check grouped rows lacking `oidc_iss` against each other, so two
gaps went uncaught: two users sharing an *identical* full `(iss, sub)` pair
would still pass the pre-flight and die on the new unique constraint
mid-migration with a bare violation instead of an actionable message; worse,
a sub-only row and a full-pair row sharing one `sub` passed cleanly with
**no failure at all** — both migrated, and the sub-only row became
permanently unreachable, since `get_by_oidc_identity` always tries the exact
`(iss, sub)` match first. Fixed by grouping every row carrying `oidc_sub`
(not only issuer-less ones). New tests:
`test_backfill_refuses_when_two_users_share_an_identical_full_pair` and, for
the case that previously failed silently,
`test_backfill_refuses_when_a_legacy_row_shares_a_subject_with_a_full_pair`.
(The first version of this fix over-corrected — see round 4, item 1, below.)

**3. The `OidcIdentityRaceError` → 503 mapping had no route-level test** —
only the store-level raise was covered. Added
`test_identity_race_contention_maps_to_503`, using a fake store that always
raises, asserting both the status code and the `Retry-After` header.

**4. Two comments cited `CLAUDE.md` by name.** This repository is public;
a comment should state the rule it invokes (a stable, disclosed pick beats a
silent or unstable one) rather than point an outside reader at an internal
file they have no reason to treat as authoritative. Reworded in
`user_store.py` and `test_user_store_oidc.py`.

## Fixes, round 4 (adversarial re-review of round 3)

**1. The round-3 pre-flight widening over-blocked a case that isn't
actually broken.** Grouping by every row carrying `oidc_sub` and refusing on
any duplicate also refused two full-pair rows sharing a subject but naming
two *different*, non-null issuers — a shape that is not ambiguous at all:
`UNIQUE(iss, sub)` is satisfied by both, the NULL-issuer partial index
applies to neither, and a later login matches exactly one of them by its
exact `(iss, sub)` pair. Telling this apart from a real collision is not
hard, and the previous version's docstring/message admitted skipping the
distinction rather than making it. Fixed by narrowing the `HAVING` clause to
the two shapes that are genuinely ambiguous: an identical `(iss, sub)` pair
repeating, or any NULL issuer present in a group of more than one subject.
A subject shared only by rows with distinct non-null issuers now passes
through untouched — which matters beyond this migration, since the tenant
work ahead is expected to make exactly that shape routine once more than one
issuer exists; an over-strict check here would have turned this migration
into something that blocks the deployments the next phase creates. Dropped
the docstring paragraph and error-message clause that had apologized for
over-blocking, since the case is no longer blocked. New test:
`test_backfill_allows_two_distinct_issuers_sharing_a_subject`, alongside the
existing refusal tests for the two shapes that still correctly block.

**Flagged, not fixed — out of scope for this PR:** every failure on this
callback (401, 409, 503) currently returns a raw FastAPI JSON error body to
a browser, because the gateway registers no exception handler for this route
and the frontend handles none of these responses either. The sensible shape
is a redirect back to the login page carrying an error the frontend can
render, and that applies to all of this endpoint's failure modes uniformly,
not just the two this PR adds (503) or touches (409) — it belongs in its own
change rather than being special-cased here for just the codes this PR
happens to touch.

## H. Flagged for a human decision — not implemented

Any account with a matching *verified* email is claimable by an OIDC login,
by design — that's the ticket. This is not a bug; it is the direct, intended
consequence of "verified email decides the account" applied without
exception. But two specific accounts deserve a deliberate decision rather
than an implicit one, because both are fixed, guessable, login-less
addresses that the system itself relies on:

- **The seeded `GATEWAY_ADMIN_EMAIL` admin**, which exists in every
  deployment. If an attacker controls (or an IdP asserts a verified claim
  for) that exact address, they get the admin account.
- **`agent-bootstrap@switch.local`**, seeded by a sibling PR (#399,
  `work/multi-tenancy-phase2-agent-reg`) as the non-admin owner for agents
  registered through the deployment-wide registration token — it exists so
  that shared token doesn't confer admin authority. Today it is protected by
  exactly the refusal this PR removes. Once this lands, whoever can make an
  IdP assert that address with `email_verified=true` becomes the *owner* of
  every agent registered through the bootstrap path — not admin, but
  owner-based authorization treats that as full management rights over those
  agents and everything they own.

These are the same class of problem, so the right shape is **a
not-linkable list of seeded system addresses**, not a special case for the
admin alone. A mitigation that protects only `GATEWAY_ADMIN_EMAIL` would
still leave the bootstrap account exposed.

**Important limitation of any email-based mitigation, including the
recommendation below:** the legacy sub-only identity path (fix A / round-2
item 1) bypasses email, `email_verified`, and even the claim's issuer
entirely — it resolves purely by matching a stored subject. This is not a
regression (main did the same) and is deliberately preserved and tested. But
it means that if either seeded account ever accumulates a legacy sub-only
identity row, no email-address-based protection closes that path: the
account would still be claimable by subject alone. Anyone adopting the
recommendation below should know it protects the *email-based* linking path
only.

One of the steering devs gave an initial instinct here, explicitly deferring
the actual call:

- **Leading option — protect a fixed list of seeded system addresses**
  (`GATEWAY_ADMIN_EMAIL` and `agent-bootstrap@switch.local` today; anything
  added later in the same category). Refuse to link an OIDC identity into any
  account on that list; those accounts stay on password login only.
  Rationale for: these are the accounts where the IdP's word alone should not
  be sufficient, since either grants broad standing authority (admin, or
  ownership of every bootstrap-registered agent) — and neither is anyone's
  daily-driver account, so excluding them costs almost nothing while leaving
  the ticket's actual use case (password signup, later Google) intact for
  ordinary users.
- **Rejected variant — refuse linking into any `role="admin"` account.**
  Broader, and looks appealing today, but doesn't survive the tenant model
  this Phase 2 work is heading toward: once every workspace has an owner,
  every workspace owner is an `admin` of *something*, and a blanket rule
  would then exclude precisely the people most likely to actually be on
  enterprise SSO. It also wouldn't cover the bootstrap account, which isn't
  an admin at all. The named-list version above has neither problem.
- **Rejected — refuse linking into any account with a password hash.** Would
  close the exposure entirely but also kills the ticket's actual use case
  (password signup, later Google), so it isn't really on the table.

Recommendation: the named-list option, but this is deliberately **not
implemented** here — it's a policy call for humans, not something to decide
unilaterally in this PR. **Whichever of this PR and #399 lands second
inherits the obligation to implement the list** (both PR bodies say so, so
neither side loses track of it).

Related, and also not addressed here: `config.py`'s docstring for
`gateway_oidc_require_email_verified` previously undersold what the flag
controls; updated in round 2 to say plainly what it does and does not affect,
including the legacy sub-only carve-out.

**Also flagged for the tenant work, not this PR:** nothing today binds an
issuer to an email domain, and there's no `UNIQUE(user_id, iss)`. Once a
second IdP exists (or a per-tenant IdP), one tenant's IdP admin could link
into another tenant's user by asserting that user's address. Worth
addressing when multi-tenant IdPs land, out of scope for this change.

## Safety properties and how they're tested

1. **A verified email may link; an unverified one never picks an existing,
   different account.** Exception, stated explicitly rather than implied: a
   pre-existing legacy identity (sub only, no issuer) continues to resolve
   and have its issuer backfilled purely by subject, without consulting
   email or `email_verified` at all — matching main's original behavior for
   that narrow, pre-CHOO-2624 case. Covered by
   `test_unverified_email_collision_with_different_identity_is_refused`, the
   route-level `test_email_collision_still_rejected_when_check_disabled`, and
   (for the legacy exception) `test_legacy_identity_without_issuer_resolves_even_if_email_changed`.
2. **An identity already bound to a different account is never re-bound.**
   Covered by `test_identity_already_bound_elsewhere_is_never_relinked` and
   the route-level `test_bound_identity_is_not_relinked_by_a_different_claimed_email`.
3. **Two subjects from the same issuer can't collapse identities in a way
   that lets one impersonate the other**, and **no two identity rows can
   ever resolve ambiguously** — enforced at the DB level for both the normal
   case (`UNIQUE(iss, sub)`) and the legacy NULL-issuer case
   (`UNIQUE(sub) WHERE iss IS NULL`), not just in application code. Covered
   by `test_two_subjects_same_issuer_can_both_link_to_one_account`,
   `test_iss_sub_unique_constraint_is_enforced_at_the_db_level`, and
   `test_null_issuer_sub_unique_constraint_is_enforced_at_the_db_level`.
4. **Password login is not weakened.** Covered by
   `test_verified_email_links_to_existing_account` and the route-level test
   of the same name.
5. **A pre-CHOO-2624 identity is not locked out or forked into a second
   account**, and **every genuinely ambiguous shape of duplicate `oidc_sub`
   metadata — legacy vs. legacy, full pair vs. identical full pair, or
   legacy vs. full pair — is refused rather than silently resolved by row
   order, while a subject shared only by rows naming distinct, non-null
   issuers (not ambiguous at all, and the shape the tenant work ahead will
   make routine) migrates through untouched.** Covered by the round-1
   legacy tests plus `test_backfill_refuses_when_two_users_share_a_legacy_subject`,
   `test_backfill_refuses_when_two_users_share_an_identical_full_pair`,
   `test_backfill_refuses_when_a_legacy_row_shares_a_subject_with_a_full_pair`,
   and `test_backfill_allows_two_distinct_issuers_sharing_a_subject`.
6. **A race between two logins resolves to one row without a 500, and
   without rolling back the caller's transaction**, and **a race that keeps
   losing fails loudly under its own error type rather than looping or being
   mistaken for a security rejection.** Covered by
   `test_concurrent_creation_of_the_same_identity_resolves_to_the_winner`,
   `test_concurrent_new_email_provisioning_resolves_to_one_account`, and
   `test_race_exhausted_twice_raises_a_distinct_contention_error`.
7. **A pre-existing case-duplicate email resolves the same way every time and
   the ambiguity is disclosed, not silently (or inconsistently) guessed.**
   The pick itself is an arbitrary-but-stable tiebreak (lowest `id`), not a
   claim about which account is "real". Covered by
   `test_pre_existing_case_duplicate_emails_resolve_deterministically_and_log`,
   confirmed non-flaky by running it 12 times after the round-3 fix (12/12
   passing; the pre-fix version had been observed failing about half the
   time).
8. **A raced login returns a usable error, not a generic 500 or a security
   rejection it doesn't deserve.** Covered at the store level by
   `test_race_exhausted_twice_raises_a_distinct_contention_error` and at the
   route level by `test_identity_race_contention_maps_to_503`.

## Test plan

- [x] `just test` (full suite: 2587 passed)
- [x] `just typecheck`
- [x] `just format` / `ruff check`
- [x] The previously-flaky case-duplicate-email test run 12 times in
      isolation after the round-3 fix: 12/12 passing.
- [x] Manual `alembic upgrade` / `downgrade` against throwaway Postgres
      instances, seven separate rehearsals in total across all four review
      rounds: (1) a full pair + a sub-only legacy row, verifying backfill and
      exact downgrade round-trip; (2) a mixed-case email row, verifying the
      case-insensitive index; (3) two users sharing one legacy subject,
      verifying the migration refuses with a clear message, leaves no partial
      schema behind, and — after resolving the collision by hand — succeeds,
      with the partial unique index then independently confirmed to reject a
      direct duplicate insert; (4) two users sharing an identical full
      `(iss, sub)` pair, verifying the widened pre-flight catches it with both
      issuers named in the message; (5) a legacy row and a full-pair row
      sharing one subject — the case that previously migrated silently and
      wrong — verifying the widened pre-flight now catches it too; (6) all
      three shapes seeded in the same database at once (identical full pair,
      legacy-vs-full-pair, and two distinct issuers sharing a subject),
      verifying the error names only the two genuine collisions and migrates
      cleanly once those two are resolved by hand — with the two-distinct-
      issuer rows landing in `oidc_identities` as two independent rows,
      untouched; (7) downgrade of that same database, confirming both
      distinct-issuer rows round-trip back into `users.metadata`
      independently.

### DIFF
--- a/core/switch_core/config.py
+++ b/core/switch_core/config.py
@@ -60,9 +60,16 @@ class SwitchConfig(BaseSettings):
     #
     # This only controls whether an unverified login is accepted at all — for
     # a brand-new identity, or one already linked to an account. It never lets
-    # an unverified email link a new identity into a *different*, pre-existing
-    # account: that always requires the IdP to assert `email_verified=true`,
-    # regardless of this setting. See UserStore.get_or_create_oidc_user.
+    # an unverified email link a *new* identity into a *different*,
+    # pre-existing account: that always requires the IdP to assert
+    # `email_verified=true`, regardless of this setting. See
+    # UserStore.get_or_create_oidc_user.
+    #
+    # Exception: a legacy identity linked before issuers were tracked at all
+    # resolves, and has its issuer backfilled, purely by matching its stored
+    # subject — it never consults email, `email_verified`, or this setting.
+    # That is not new here; it is the same subject-only match this login has
+    # always done for such a row. See OidcIdentity's docstring in models.py.
     gateway_oidc_require_email_verified: bool = True
     # Lets the password login path be disabled (OIDC-only) without code changes.
     gateway_password_login_enabled: bool = True
--- a/core/switch_core/config.py
+++ b/core/switch_core/config.py
@@ -57,6 +57,12 @@ class SwitchConfig(BaseSettings):
     # directory-provisioned users are permanently false and cannot be fixed
     # from the Okta side. Set false ONLY for a single-tenant IdP whose
     # addresses are authoritative (corporate directory, HR-provisioned).
+    #
+    # This only controls whether an unverified login is accepted at all — for
+    # a brand-new identity, or one already linked to an account. It never lets
+    # an unverified email link a new identity into a *different*, pre-existing
+    # account: that always requires the IdP to assert `email_verified=true`,
+    # regardless of this setting. See UserStore.get_or_create_oidc_user.
     gateway_oidc_require_email_verified: bool = True
     # Lets the password login path be disabled (OIDC-only) without code changes.
     gateway_password_login_enabled: bool = True
--- a/core/switch_core/db/models.py
+++ b/core/switch_core/db/models.py
@@ -71,15 +71,34 @@ class OidcIdentity(Base):
     predates the issuer being tracked at all (``oidc_sub`` stored alone); the
     application never writes a NULL issuer itself. Such a row still matches on
     ``sub`` alone and has its issuer backfilled on the next login, same as
-    before this identity had its own table — ``sub`` is unique per issuer, so
-    matching on it alone is safe in practice given a deployment registers one
-    issuer.
+    before this identity had its own table. ``UNIQUE(iss, sub)`` does not
+    constrain these rows at all — Postgres treats every NULL as distinct from
+    every other NULL — so ``ix_oidc_identities_sub_null_iss`` separately
+    enforces at most one NULL-issuer row per ``sub``; without it two different
+    users could each hold one for the same ``sub``, and which one a login
+    resolved to would depend on row order, silently annexing one account and
+    orphaning the other.
+
+    Matching a legacy row on ``sub`` alone, with no issuer or email in the
+    comparison at all, is only safe because a deployment has exactly one
+    configured issuer (``gateway_oidc_issuer_url``) today. The moment a second
+    issuer exists — a later multi-tenancy phase — two different real-world
+    identities could share the same subject string, and a legacy row would
+    resolve to whichever one asserts it first regardless of which issuer
+    actually owns it. Re-verify or purge sub-only rows before a deployment is
+    ever configured with more than one issuer.
     """
 
     __tablename__ = "oidc_identities"
     __table_args__ = (
         UniqueConstraint("iss", "sub", name="uq_oidc_identities_iss_sub"),
         Index("ix_oidc_identities_sub", "sub"),
+        Index(
+            "ix_oidc_identities_sub_null_iss",
+            "sub",
+            unique=True,
+            postgresql_where=text("iss IS NULL"),
+        ),
     )
 
     id: Mapped[str] = mapped_column(Text, primary_key=True, default=_uuid)
--- a/core/switch_core/db/models.py
+++ b/core/switch_core/db/models.py
@@ -51,6 +51,13 @@ class User(Base):
     )
 
 
+# An OIDC login matches an account on email case-insensitively (CHOO-2624):
+# the IdP and the person typing a password don't reliably agree on casing for
+# the same mailbox, so a case-sensitive compare would silently defeat the
+# "same account" guarantee. Backs UserStore.get_by_email's lower() compare.
+Index("ix_users_email_lower", func.lower(User.email))
+
+
 class OidcIdentity(Base):
     """A verified IdP identity linked to a user (CHOO-2624).
 
@@ -59,16 +66,25 @@ class OidcIdentity(Base):
     verified email, not on login method, so a password sign-up that later
     signs in with an IdP sharing its email gets this identity added to the
     same account rather than a second one.
+
+    ``iss`` is nullable only for rows migrated from a pre-CHOO-2624 login that
+    predates the issuer being tracked at all (``oidc_sub`` stored alone); the
+    application never writes a NULL issuer itself. Such a row still matches on
+    ``sub`` alone and has its issuer backfilled on the next login, same as
+    before this identity had its own table — ``sub`` is unique per issuer, so
+    matching on it alone is safe in practice given a deployment registers one
+    issuer.
     """
 
     __tablename__ = "oidc_identities"
     __table_args__ = (
         UniqueConstraint("iss", "sub", name="uq_oidc_identities_iss_sub"),
+        Index("ix_oidc_identities_sub", "sub"),
     )
 
     id: Mapped[str] = mapped_column(Text, primary_key=True, default=_uuid)
     user_id: Mapped[str] = mapped_column(Text, ForeignKey("users.id"), nullable=False)
-    iss: Mapped[str] = mapped_column(Text, nullable=False)
+    iss: Mapped[str | None] = mapped_column(Text, nullable=True)
     sub: Mapped[str] = mapped_column(Text, nullable=False)
     created_at: Mapped[str] = mapped_column(
         DateTime(timezone=True), server_default=func.now(), nullable=False
--- a/core/switch_core/db/models.py
+++ b/core/switch_core/db/models.py
@@ -51,6 +51,30 @@ class User(Base):
     )
 
 
+class OidcIdentity(Base):
+    """A verified IdP identity linked to a user (CHOO-2624).
+
+    One row per linked ``(iss, sub)``, unique so a subject can never bind to
+    more than one account. A user may hold several — accounts are keyed on
+    verified email, not on login method, so a password sign-up that later
+    signs in with an IdP sharing its email gets this identity added to the
+    same account rather than a second one.
+    """
+
+    __tablename__ = "oidc_identities"
+    __table_args__ = (
+        UniqueConstraint("iss", "sub", name="uq_oidc_identities_iss_sub"),
+    )
+
+    id: Mapped[str] = mapped_column(Text, primary_key=True, default=_uuid)
+    user_id: Mapped[str] = mapped_column(Text, ForeignKey("users.id"), nullable=False)
+    iss: Mapped[str] = mapped_column(Text, nullable=False)
+    sub: Mapped[str] = mapped_column(Text, nullable=False)
+    created_at: Mapped[str] = mapped_column(
+        DateTime(timezone=True), server_default=func.now(), nullable=False
+    )
+
+
 # ── API Keys ─────────────────────────────────────────────────────────────────
 
 
--- a/core/switch_core/db/stores/user_store.py
+++ b/core/switch_core/db/stores/user_store.py
@@ -3,14 +3,16 @@ from __future__ import annotations
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from switch_core.db.models import User
+from switch_core.db.models import OidcIdentity, User
 
 
 class OidcIdentityConflictError(Exception):
-    """An OIDC login's email already belongs to a different local account.
+    """An unverified OIDC login's email already belongs to a different account.
 
     Raised instead of silently linking the IdP identity to a pre-existing
-    account: auto-linking by email is an account-takeover vector.
+    account: an unverified email is attacker-controllable, so trusting it to
+    pick an account is an account-takeover vector. A verified email links
+    instead — see ``get_or_create_oidc_user``.
     """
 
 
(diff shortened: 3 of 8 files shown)


wojtyniakAQ:
recheck

amaudruz (approved):

PR #398, opened 2026-09-08
fix(console): manual refresh re-checks sign-in options

wojtyniakAQ:
## What the user saw

A Switch server that gained OIDC sign-in (operator configures an IdP and
restarts the gateway) kept showing password-only in Switch Console until the
whole app was restarted. Clicking Refresh on the server page did not help —
verified live against a server returning
`{"password_login_enabled":true,"oidc_enabled":true,"oidc_provider_label":"WorkOS"}`
while Console kept offering only email/password.

## Why Refresh didn't work

`refreshServer` — the server page's manual re-check — called `refreshStatus`
and `ensureAuthConfig`. `ensureAuthConfig` fetches a server's login methods
once and caches them for the life of the process
(`if (this.authConfigs.has(serverId)) return;`), which is right for its
other callers (mount, recovery sweeps) but meant the explicit user-initiated
refresh could never see a config change after the first successful fetch.

## The fix

Split the actual fetch (in-flight collapsing, host-blocked skip, the
try/catch that marks a server unreachable without raising the global error
banner) into a private `fetchAuthConfig`. `ensureAuthConfig` keeps its
cache-once behavior; a new `refreshAuthConfig` skips the cache check, and
`refreshServer` calls it. Every other caller of `ensureAuthConfig`
(`useServerSignIn`'s mount effect, the fan-out around `view.tsx`) stays
exactly as cheap as before.

## Round 2: a race in the reachability flag

`refreshServer` runs the connection-status check and the sign-in-options
check concurrently. The first version of this fix had both write the same
`unreachable` `Set` with no coordination, so whichever settled last decided
the page's reachability regardless of what the other found — a signed-in
server whose sign-in-options endpoint blipped could get thrown onto the
"cannot reach the server" screen even though its session was fine, and this
now recurred on every Refresh click and every 10-second auto-retry tick
instead of the narrow mount-time window it used to be confined to.

Fixed by tracking the two checks in separate sets, `statusUnreachable` and
`authConfigUnreachable`, with `isUnreachable` deferring to status once a
status read has ever run for a server.

That version also introduced `retryConnection`, used by the unreachable
card's 10-second timer, which re-checks connectivity but keeps sign-in
options at their once-cached read — so an outage does not turn into an
unbounded repeated sign-in-options fetch.

## Round 3: the fix above deleted a signal instead of separating it

Review caught that `refreshStatus` writes `statuses` on both success and
failure, and `init()`/`addServer` populate it for every server within the
first tick of the app's life — so `isUnreachable`'s fallback to
`authConfigUnreachable` was live for a fraction of a second and dead for the
rest of the session. The consequence was not the transient hiccup this PR
set out to fix but the durable one: if the sign-in-options endpoint stayed
persistently broken, there was no signal at all — no unreachable card,
since status was fine, and no error banner by design — so the panel kept
rendering the last cached options as current, forever, silently.

Fixed by adding `authConfigCheckFailed(serverId)`, a signal kept separate
from page-level reachability (`isUnreachable` still defers to status alone,
unchanged from round 2). The sign-in panel (`server-sign-in.tsx`) reads it
and shows "Could not check for updated sign-in options — showing what was
last known" alongside the (possibly stale) cached fields, or "Could not
check sign-in options." if nothing was ever cached — disclosure at the
point the stale data is actually shown, not at the page level.

The same review flagged that the header's "Refresh" button and the
unreachable card's "Retry" button looked identical but behaved differently
(only the header forced a fresh sign-in-options check). Resolved by
collapsing the distinction rather than making it legible: both buttons are
single, bounded, explicit clicks, so both now call the full `refreshServer`.
Only the card's own repeating 10-second timer keeps the cheaper
`retryConnection`, since it is the one thing here that fires without a
click and must stay cheap forever.

## Round 4: two edges in the disclosure itself

1. **The failure text could show while a retry was in flight.** The panel
   chose its "Checking sign-in options…" vs "Could not check sign-in
   options." text from `configCheckFailed` alone. A first attempt failing
   with nothing cached, followed by a remount that starts a fresh fetch,
   would claim the check had failed right up until that fetch actually
   settled — true only briefly, but false while it lasted. Added
   `authConfigChecking(serverId)`, exposed the same way as
   `authConfigCheckFailed`, and the panel now prefers the checking state
   whenever a fetch is genuinely running.
2. **A host-blocked skip left a stale failure flag behind.**
   `fetchAuthConfig`'s host-blocked early return exited without touching
   `authConfigUnreachable` — correct in that it didn't fabricate a new
   failure, but a failure recorded just before the host went down would
   survive the outage and reappear once unblocked, since the recovering
   `ensureAuthConfig` call no-ops on a config that doesn't need fetching and
   never clears it. Now the skip clears `authConfigUnreachable`: a failure
   from before the block belongs to a check that isn't running, so it
   shouldn't outlive it.

## The stale-config trade-off

On a failed re-fetch, the fix leaves the previously cached config in place
rather than dropping it, and now discloses the failure in the one place
that matters (the sign-in panel) rather than at the page level, so a
reachable server never gets evicted over its own sign-in-options endpoint
having a bad day — durably or transiently.

## Test plan

- [x] Added a test that caches a config, has the RPC return a different one,
      calls `refreshServer`, and asserts the store now exposes the new
      config.
- [x] Added a test asserting `retryConnection` does not re-fetch a cached
      config, only connectivity.
- [x] Added a test asserting a sign-in-options failure that lands after a
      successful status check does not flip `isUnreachable` to true (the
      round-2 race).
- [x] Added tests for the round-3 defect: a persistently failing
      sign-in-options check stays disclosed via `authConfigCheckFailed`
      without evicting an otherwise-reachable server, and the disclosure
      clears once the endpoint answers again.
- [x] Added tests for the round-4 edges: a retry in flight reports
      `authConfigChecking` true and is not simultaneously reported as
      failed; a stale failure recorded before a host-blocked skip is
      cleared by that skip rather than surviving it. Both fail against the
      prior commit.
- [x] `pnpm vitest run --project node src/renderer/features/switch-servers/switch-servers-store.test.ts` — 31 passed
- [x] `pnpm vitest run --project node --project main-db` — full suite passed except one pre-existing failure in `src/sidecar/session-spawner.test.ts` unrelated to this change (reproduces identically on unmodified `main`)
- [x] `pnpm run typecheck` (workspace) — passed
- [x] `pnpm run lint` (workspace) — passed, one pre-existing unrelated warning
- [x] `pnpm run format:check` (workspace) — passed

### DIFF
--- a/console/apps/switch-console-desktop/src/renderer/features/switch-servers/server-sign-in.tsx
+++ b/console/apps/switch-console-desktop/src/renderer/features/switch-servers/server-sign-in.tsx
@@ -1,3 +1,4 @@
+import { TriangleAlert } from 'lucide-react';
 import { observer } from 'mobx-react-lite';
 import { useEffect, useState } from 'react';
 import { Button } from '@renderer/lib/ui/button';
@@ -18,6 +19,11 @@ import { switchServersStore } from './switch-servers-store';
 export type ServerSignIn = {
   /** Which methods the server offers, or null while that is still unknown. */
   config: SwitchAuthConfig | null;
+  /** Whether the last attempt to read sign-in options failed. The page never
+   * lets this evict a reachable server (see `isUnreachable` on the store), so
+   * this is the only place that failure is visible — `config` may be a stale
+   * answer from before the endpoint broke, or null if it never succeeded. */
+  configCheckFailed: boolean;
   email: string;
   password: string;
   setEmail: (value: string) => void;
@@ -56,6 +62,7 @@ export function useServerSignIn(serverId: string): ServerSignIn {
 
   return {
     config: switchServersStore.authConfigFor(serverId),
+    configCheckFailed: switchServersStore.authConfigCheckFailed(serverId),
     email,
     password,
     setEmail,
@@ -107,11 +114,24 @@ export const ServerSignInFields = observer(function ServerSignInFields({
   };
 
   if (!config) {
-    return <p className="text-sm text-foreground-muted">Checking sign-in options…</p>;
+    return (
+      <p className="text-sm text-foreground-muted">
+        {signIn.configCheckFailed
+          ? 'Could not check sign-in options.'
+          : 'Checking sign-in options…'}
+      </p>
+    );
   }
 
   return (
     <div className="flex flex-col gap-4">
+      {signIn.configCheckFailed && (
+        <p className="flex items-center gap-1 text-xs text-amber-600 dark:text-amber-500">
+          <TriangleAlert className="size-3 shrink-0" />
+          Could not check for updated sign-in options — showing what was last known.
+        </p>
+      )}
+
       {config.passwordLoginEnabled && (
         <div className="flex flex-col gap-3">
           <div className="space-y-1.5">
--- a/console/apps/switch-console-desktop/src/renderer/features/switch-servers/server-sign-in.tsx
+++ b/console/apps/switch-console-desktop/src/renderer/features/switch-servers/server-sign-in.tsx
@@ -24,6 +24,10 @@ export type ServerSignIn = {
    * this is the only place that failure is visible — `config` may be a stale
    * answer from before the endpoint broke, or null if it never succeeded. */
   configCheckFailed: boolean;
+  /** Whether a sign-in-options read is in flight right now — checked before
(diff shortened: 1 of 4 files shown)


wojtyniakAQ:
recheck

amaudruz (approved):

PR #399, opened 2026-09-08
fix(agent): scope AGENT_REGISTRATION_TOKEN to a non-admin owner

wojtyniakAQ:
## The attack this closes

`AGENT_REGISTRATION_TOKEN` is seeded once at startup as an `ApiKey` of type
`"registration"`, owned by the seeded admin user. Every agent registered
through `POST /agents` with that token gets `owner_id = <admin's id>`.

Room and resource authorization (`ProtocolService._resolve_acting_identity`,
used by room creation, roster changes, and reference/package attachment)
looks up the agent's owner and treats `owner.role == "admin"` as acting with
full admin authority — the same global bypass `authz.can`/`can_manage` give a
human admin. So an agent registered through the shared deployment token
doesn't just get an owner; it gets **admin authority over every room and
resource in the deployment**, not just ones it's a member of.

This token is explicitly meant to be shared: `docs/official/deploy/self-host.md`
says "Set `AGENT_REGISTRATION_TOKEN` before the first start... which is what
your colleagues' agents register against." So the documented, intended
workflow — hand the one deployment secret to each colleague so they can
bring up their own agent — was itself the privilege-escalation path: any
colleague who registers an agent this way ends up with an agent that can
bypass every room's `read_visibility`/`write_visibility`, change any room's
membership, and manage any reference/document/package, because it is
"owned" by an admin.

A secondary, related gap in the same code path: `ProtocolService.
register_agent_with_token` (used by server-side connectors) never checked
the presented key's `type`, so a leaked **agent** API key — far more widely
distributed than a registration token, since every live connection uses one
— could also be replayed to register more agents under that agent owner's
identity.

## The fix

`AGENT_REGISTRATION_TOKEN` remains (a fresh deployment still needs some way
to register its first agents before anyone has logged into the gateway), but
it is seeded as a new, distinct `ApiKey` type, `"bootstrap"`, rather than
`"registration"`. Agents registered through a bootstrap-type key are owned by
a dedicated, non-admin account (`agent-bootstrap@switch.local`, seeded
alongside it) instead of the admin — so holding the shared secret no longer
confers admin authority, while the resulting agent still has a real owner and
can still create rooms, own them, and attach references, same as before.

The `ApiKey` row itself stays filed under the admin user, so it keeps showing
up (and can be deleted) on the existing gateway "API Keys" page — no new UI
plumbing, just a one-line filter update so a `"bootstrap"`- or `"retired"`-
type key (see "Round 4" below) isn't hidden. Two things that page's existing
delete button did not previously guarantee, now hold:

- **Revocation is durable.** Before this change, deleting the seeded key from
  the API Keys page didn't stick — the next restart matched the current
  `AGENT_REGISTRATION_TOKEN` value against existing hashes, found none, and
  silently reseeded the exact same key. A revocation now persists across a
  restart with the same token, and forever — even through a later rotation
  away and back (see "Round 3" below); rotating the token to a *different,
  never-used* value re-establishes bootstrap registration.
- **A leaked agent key can't be replayed as a registration token** — both
  `register_agent_with_token` and the HTTP registration dependency now
  reject any key whose type isn't `"registration"` or `"bootstrap"`.

An existing deployment's already-seeded legacy key (type `"registration"`,
label `"Default (from AGENT_REGISTRATION_TOKEN)"`) is migrated onto the new
type automatically on the next restart if its hash still matches the
configured token, and retired in place (see "Round 4") if it doesn't — no
operator action required either way.

Per-user registration keys minted from the gateway (`POST /gateway/api-keys`)
are untouched — they already scope to the minting user's own id, which is
exactly the "per-tenant credential" shape CHOO-2624 asks for; `owner_id` was
already the field a future tenant model would replace this with, so nothing
here stands in the way of that later change.

## Adversarial review, round 2

An independent pass found five real gaps in the first version of this fix,
plus a comment that narrated the change instead of documenting behavior:

1. **Changing `GATEWAY_ADMIN_EMAIL` on a running deployment either crashed
   startup or silently un-revoked a revoked key.** The first version decided
   "does a bootstrap key already exist" and stored the revocation marker
   scoped to whichever user the *currently configured* admin email resolves
   to. Point that config at a different account (token unchanged) and the
   seeder found no key under the "new" admin and tried to insert one with a
   hash the old admin's row already held — `api_keys.key_hash` is unique, so
   this failed the whole boot. The same scenario with a previously revoked
   key silently recreated the deployment-wide credential instead, logged at
   `info`. Both existence and rotation are now resolved by the key's own
   hash/type **globally**, never by which admin the config currently names —
   an admin row is only looked up to own a freshly *created* key.
2. **The bootstrap owner's identity was adopted, not verified** (superseded
   by round 3, then round 4, below).
3. **Interaction with #397, and this needs resolving between the two PRs.**
   #397 changes OIDC login so a *verified* email match to an existing account
   links to it instead of refusing. Today, the refusal is what protects
   `agent-bootstrap@switch.local` (and the seeded admin address) from being
   claimed by an external IdP identity. Once #397 lands, an IdP identity
   asserting a verified `agent-bootstrap@switch.local` would be linked to the
   account this PR seeds — handing an external party control of the shared
   identity every bootstrap-registered agent is owned by. #397's own
   write-up discusses the seeded *admin* address and does not mention this
   one. **Whichever of the two PRs lands second must add a not-linkable list
   covering both seeded addresses** (admin and bootstrap-owner) before OIDC
   linking-on-verified-email is safe to ship alongside this.
4. **Legacy-key migration matched on a free-text label alone** (refined by
   round 3, then round 4, below).
5. **Revocation was a one-way boolean** (superseded by round 3, below).
6. **Admin-owned agents predating this fix were mentioned only in this PR
   body.** Seeding now logs a `logger.warning` naming every agent still owned
   by the admin, on every startup for as long as any exist.

## Adversarial review, round 3

A second independent pass found that two of round 2's own fixes reopened
real gaps, one that round 2 explicitly protected against, plus four smaller
issues:

1. **The round-2 hash-match requirement for legacy migration orphaned a live
   admin-authority key on an ordinary upgrade path.** An admin who rotates
   `AGENT_REGISTRATION_TOKEN` at or before the upgrade — normal practice, and
   what re-running `just init-env` does — leaves the legacy row's hash stale.
   A hash-only lookup for the *current* token never finds it, so a fresh
   bootstrap key gets created at the new value while the old row is left
   completely untouched: still `type="registration"`, still owned by the
   admin, still authenticating anyone holding the old token as the admin.
   This is permanent (later restarts stop looking once a bootstrap key
   exists) and was invisible (the round-2 startup warning enumerates
   admin-owned *agents*, not stale admin-owned registration *keys*). Every
   row carrying the legacy label is now inspected regardless of hash: one
   matching the current token still migrates in place (keeping round 2's
   protection against clobbering an unrelated personal key that happens to
   share the label into becoming the live singleton); every other one is
   neutralized (deleted at the time — refined to *retired in place* by round
   4, below) with a `logger.warning` naming it, since a hash mismatch on this
   auto-generated label means the row is stale and must stop authenticating.
2. **The role check was the wrong shape and missed the realistic threat.** It
   refused only `role == "admin"`, but OIDC JIT provisioning always creates
   `role="user"` (hardcoded in `user_store.py`) — so an identity that
   squatted `agent-bootstrap@switch.local` on an older deployment, before
   this ever ran, would be silently adopted exactly as before, and could log
   in and manage every bootstrap-registered agent. The account this module
   creates is now stamped with a marker in its own `metadata_` at creation
   time; both `ensure_bootstrap_owner` (startup) and
   `resolve_registration_owner_id` (every bootstrap-token registration) now
   refuse to adopt an existing account that lacks it, regardless of role —
   this subsumes the admin-role case, which stays as a second, independent
   check (refined by round 4, below, to also accept a *genuine* pre-marker
   account rather than refusing it).
3. **Changing the revocation marker from a boolean to "last seeded hash"
   reopened the case the boolean covered.** Revoke at X (marker becomes
   `H(X)`), rotate to Y and restart (marker overwritten to `H(Y)`), rotate
   back to X and restart: the marker no longer remembers X was ever revoked,
   so the deployment silently starts serving the exact value the operator
   deliberately killed. A permanent list of every revoked hash is now kept
   instead of just the last one, and any of them is refused forever — as a
   fresh seed target and as a rotation target — no matter how many rotations
   happen in between.
4. **The startup refusal named a remedy that does not exist.** "Demote or
   remove it" implied a gateway action; there is no update-role or
   delete-user endpoint, and the gateway is down at startup regardless. Both
   messages now say plainly that the fix is a direct edit to the `users`
   table.
5. Four smaller ones, all fixed: a `get_by_type` lookup whose docstring
   asserted a system-wide singleton nothing enforced now raises if more than
   one row is ever found; a missing/malformed bootstrap owner surfaced as a
   bare 500 on the HTTP registration path with the reason only in the log
   (refined by round 4, below, to a generic 503); legacy migration skipped
   refreshing the encrypted blob, so a `JWT_SECRET_KEY` rotation alongside a
   migrated key would have broken the reveal button; and a test asserting
   only that *some* key existed after an admin-email change now asserts
   exactly one bootstrap key exists with the expected hash.

## Adversarial review, round 4

A third independent pass found that two of round 3's own fixes had gaps,
plus one that the round-2/3 admin-role tests weren't actually exercising,
plus three smaller issues:

1. **The marker check (round 3, point 2) bricked any database seeded by an
   earlier commit of this branch.** `_raise_unless_genuine` tested the
   marker before the role, so a bootstrap owner created by any earlier
   commit of this same feature has no marker, and the next boot refused to
   start — telling the operator (in practice: a reviewer, or whoever had
   already run the branch) the account "was not created by agent-
   registration bootstrap seeding", which is false. An unmarked row is now
   accepted, and backfilled with the marker, if it carries the
   last-seeded-hash key — a value only this module's own seeding step ever
   writes, so a squatter cannot have it, which is what makes "predates the
   marker" distinguishable from "put there by something else" without
   weakening the check.
2. **A retired stale key's hash was never added to the revoked list** — one
   line missing from the round-3 deletion loop. Reproduces, inside the code
   meant to prevent it, the exact bug the revoked-hashes list exists to
   close: delete the stale row at `H(old)`, then an operator restores an old
   `.env` or redeploys a stale values file setting the token back to `old`,
   and the rotation path accepts it because `H(old)` was never recorded (in
   testing, it instead crashed the boot on the `key_hash` unique constraint,
   colliding with the still-present retired row — worse, not better). Its
   hash is now added to the permanently-revoked list at the moment it is
   retired.
3. **Retiring a stale legacy key should retire it, not delete it (round 3,
   point 1).** The row must stop authenticating, but deletion isn't the
   minimal way to achieve that: every consumer already refuses any type
   outside `REGISTRATION_KEY_TYPES`, so retyping it to a new, excluded
   `"retired"` type 401s it exactly as deletion would. Deletion cost three
   things retyping doesn't: the API Keys page filters on type, so a retired
   row stays visible and hand-deletable where a deleted one leaves nothing
   but a startup log line; there is no `updated_at` and no audit table, so
   nothing else records a deleted row ever existed; and `get_by_label` is
   global while any authenticated user can create a registration key with a
   free-text label, so a colleague who happened to reuse the legacy label
   had their key destroyed at the next restart with no notification besides
   a log line the operator may never read. Because the two cases (a stale
   bootstrap key vs. a coincidentally-labeled personal key) still can't be
   told apart at seed time, the fix now preserves the evidence that lets a
   human tell them apart afterward: retype to `"retired"`, rewrite the label
   to say so, keep the warning, leave the row for the operator to review and
   delete themselves.
4. **Two tests exercising the admin-role squatter used an unmarked fixture**,
   so both actually hit the (separately tested) missing-marker check and
   never reached the role check — they passed only because `match="admin"`
   happened to match "admin authority" in the missing-marker message too.
   Both fixtures now carry the marker, so they exercise the role check they
   claim to, and match a message unique to it.
5. **The HTTP registration path returned the internal `RuntimeError` text
   verbatim** — the account name and the remediation procedure, and
   confirmation to a squatter that their squat was detected. It now logs the
   detail and returns a generic 503 (a misconfigured deployment, not a
   client error). The service-layer registration path
   (`register_agent_with_token`, used by server-side connectors) gets the
   same log-detail-return-generic treatment for consistency between the two
   registration surfaces.
6. Two smaller ones: the revoked-hashes metadata value was read with a bare
   `list(...)`, so a hand-edited non-list value (e.g. a string) would be
   silently iterated character-by-character into "nothing is revoked" —
   guarded with an `isinstance` check that raises instead; and the
   legacy-migration and stale-retirement logic are now scoped identically
   (both only run while no bootstrap key exists yet) rather than the
   deletion loop being reachable slightly more broadly than the migration
   check for no functional reason.

## Adversarial review, round 5

A fourth independent pass, focused on attacking the retention-vs-revocation
design from round 4, confirmed it holds — the revoked-hashes list lives on
the bootstrap owner's row rather than being derived from the key table, so
it survives both the retired row's own later deletion and the constraint
that happens to also guard the case where that row is still present — and
found three small gaps:

1. **No test covered the case with no constraint to fall back on.** There
   was a test for restoring the old token while the retired row is still
   present (blocked twice over: revoked-hashes, and the unique constraint on
   `key_hash`, since the retired row still holds the old value). There was
   none for deleting that retired row first — exactly what its own warning
   invites an operator to do — and *then* restoring the old token, the one
   path where the revoked-hashes list is the only thing standing in the way,
   with nothing left to collide with. Added.
2. **The missing-genuine-account message claimed more than it knows.** The
   backfill (round 4) rescues a database seeded by the third commit of this
   branch, because that commit wrote the last-seeded-hash to the bootstrap
   owner's own row. The first two commits wrote their state to the *admin's*
   row instead, so their bootstrap owner has no evidence on it and is
   correctly left unrescued — but the refusal said the account "was not
   created by agent-registration bootstrap seeding", which is not something
   this code can actually tell apart from "created by an earlier revision
   that recorded its state elsewhere." The message now names both
   possibilities instead of asserting the first as fact.
3. **`register_agent_with_token` reused `PermissionError` for a
   configuration fault.** A caller reasonably reads `PermissionError` as
   "bad token," so a server-side connector would report an authentication
   failure when the real cause is a misconfigured bootstrap owner. It now
   raises a dedicated `BootstrapOwnerMisconfiguredError`; the generic 503 on
   the HTTP registration path was already right and is unchanged.

Two smaller things worth calling out rather than changing here:

- Filing the bootstrap `ApiKey` under the admin (for visibility/revocation on
  the existing page) but attributing agents it registers to the bootstrap
  owner means a bootstrap-registered agent's *own* `"agent"`-type key is now
  filed under a login-less account. The admin can still delete the whole
  agent (and its key with it) from the Agents page, but can no longer reveal
  or delete just that key from the API Keys page the way they could when it
  was filed under their own account. This is a real, if narrow, behavior
  change from before this PR.
- The Registration Keys page shows no `type` column and no distinction
  between an ordinary key, the deployment bootstrap key, and a retired one.
  A retired key's rewritten label explains why it stopped working, which is
  adequate, but its Copy button still hands an operator a credential that no
  longer authenticates, and there's no visual cue that it's now safe to
  delete. A `type` column, or suppressing Copy for a non-`"registration"`
  type, would finish the job properly — frontend work, and belongs in its
  own change.

## Left for the tenant ticket (CHOO-2623)

- No `tenants` table or `tenant_id` column was added, per the ticket scope.
  `owner_id` (on `Agent`, `Room`, `Reference`, `Document`, `Package`) remains
  the scoping field; swapping it for a tenant id later is a single-column
  change plus updating `registration_bootstrap.py`'s one owner-resolution
  function, not a redesign.
- Agents **already registered** under the admin's real account before this
  fix retain admin-equivalent authority — this closes the vulnerability going
  forward, migrates or retires the credential, and now warns on every startup
  naming both the affected agents and any stale legacy key, but it can't
  safely distinguish "registered via the shared token" from "the admin's own
  legitimate agent" after the fact to auto-reassign existing rows. An
  operator who has shared the deployment token needs to reassign or
  re-register any agent named in that warning that shouldn't have admin
  authority.
- `docs/official/deploy/self-host.md` is generated from a separate docs repo
  (per this repo's `docs/README.md`) and still says the token is "what your
  colleagues' agents register against" implying admin ownership — that
  wording needs a follow-up fix at the source now that colleague-registered
  agents land on a dedicated non-admin account instead.
- The #397 not-linkable-list coordination described above.

## Testing

- `core/tests/switch_core/bridges/agent/test_registration_bootstrap.py` —
  the owner-resolution helper (registration vs. bootstrap vs. missing owner);
  the adversarial cases (an admin account, and separately a `role="user"`
  account, squatting the bootstrap owner's address; a bootstrap owner
  promoted to admin after the fact; a marker-less row backfilled rather than
  refused on the upgrade path).
- `core/tests/switch_core/test_main.py` — the seeding lifecycle: fresh
  install, idempotent rerun, legacy-key migration, a legacy-labeled key
  rotated before the upgrade (retired in place, still visible, no longer
  authenticates, and its old value can't be resurrected by restoring an old
  `.env`), a personal key sharing the legacy label but not its hash (retired
  and warned about, deliberately), revocation surviving a restart with the
  same token (including right after a legacy migration), rotation after
  revocation re-enabling the key, rotating *back* to a revoked value *not*
  reinstating it, the admin-owned-agents warning, a `role="user"` squatter at
  the bootstrap address blocking seeding (and, separately, an admin-role one
  that carries the marker so it actually exercises the role check), a
  marker-less-but-genuine row healed rather than refused, a malformed
  revoked-hashes value failing loud, deleting a retired key and then
  restoring its old value staying blocked with no unique-constraint backstop
  to lean on, and — the round-2 merge-blocking finding — an admin-email
  change neither crashing startup nor un-revoking a previously revoked key.
- `core/tests/switch_core/bridges/agent/protocol/test_register_with_token_bootstrap.py`
  — proves a bootstrap-token registration lands on the non-admin owner (not
  admin), that a personal registration key is unaffected, and that an agent
  key can no longer be replayed as a registration token.
- `core/tests/switch_core/bridges/agent/test_bearer_auth_registration_pass_through.py`
  — the bearer-middleware gate accepts both key types on ordinary paths and
  rejects both on `/mcp`.
- `core/tests/integration/test_registration_bootstrap_scope.py` — end-to-end
  through real Postgres: registering through `AGENT_REGISTRATION_TOKEN` (via
  the same seeding path `main.py` uses) never lands on the admin owner, and a
  bootstrap-owned agent is actually denied (`PermissionError` from
  `_require_room_action`) write access to a private room owned by the admin
  that an admin-owned agent can access — the real authorization decision this
  PR closes, not just an attribute check.
- Every new test in rounds 2 through 5 was checked against a regression: the
  fix it targets was reverted locally, the test was confirmed to fail
  (concretely, not just "some exception"), and the fix was restored.
- `just test` (2600 passed), `just typecheck`, and `just format` are clean.
  `just test-integration` passes aside from a pre-existing, unrelated flake
  in `test_agent_greeting_e2e.py` (room-join delivery timing under this
  sandbox's colima + testcontainers setup) reproduced on unmodified `main`.

No Alembic migration: `ApiKey.type` is a free-text column with no `CHECK`
constraint, so the new `"bootstrap"`/`"retired"` values need no schema change.

### DIFF
--- a/core/switch_core/bridges/agent/api/handlers.py
+++ b/core/switch_core/bridges/agent/api/handlers.py
@@ -151,7 +151,10 @@ async def _resolve_registration_user_id(
     key = await api_key_store.get_by_hash(session, token_hash)
     if key is None or key.type not in REGISTRATION_KEY_TYPES:
         raise HTTPException(status_code=401, detail="Invalid registration token")
-    return await resolve_registration_owner_id(session, protocol.user_store, key)
+    try:
+        return await resolve_registration_owner_id(session, protocol.user_store, key)
+    except RuntimeError as exc:
+        raise HTTPException(status_code=500, detail=str(exc)) from exc
 
 
 # Registration endpoints
--- a/core/switch_core/bridges/agent/api/handlers.py
+++ b/core/switch_core/bridges/agent/api/handlers.py
@@ -90,6 +90,10 @@ from switch_core.bridges.agent.protocol.connections import (
 )
 from switch_core.bridges.agent.protocol.service import AgentExistsError, ProtocolService
 from switch_core.bridges.agent.protocol.stream import event_stream
+from switch_core.bridges.agent.registration_bootstrap import (
+    REGISTRATION_KEY_TYPES,
+    resolve_registration_owner_id,
+)
 from switch_core.db.models import Agent, Task
 from switch_core.db.stores.api_key_store import ApiKeyStore
 from switch_core.db.stores.feature_flag_store import FeatureFlagStore
@@ -130,17 +134,24 @@ async def _resolve_registration_user_id(
     authorization: Annotated[str, Header()],
     session: Annotated[AsyncSession, Depends(get_session)],
     api_key_store: Annotated[ApiKeyStore, Depends(get_api_key_store)],
+    protocol: Annotated[ProtocolService, Depends(get_protocol)],
 ) -> str:
     """Validate the registration token in the Authorization header and
-    return the owning user_id."""
+    return the user_id new agents should be owned by.
+
+    A personal ``"registration"`` key resolves to the user who minted it. The
+    deployment-wide ``"bootstrap"`` key (see ``registration_bootstrap.py``)
+    resolves to a dedicated, non-admin account instead of the admin who
+    seeded it, so holding it never confers admin authority.
+    """
     if not authorization.startswith("Bearer "):
         raise HTTPException(status_code=401, detail="Invalid authorization header")
     token = authorization[7:]
     token_hash = hashlib.sha256(token.encode()).hexdigest()
     key = await api_key_store.get_by_hash(session, token_hash)
-    if key is None or key.type != "registration":
+    if key is None or key.type not in REGISTRATION_KEY_TYPES:
         raise HTTPException(status_code=401, detail="Invalid registration token")
-    return key.user_id
+    return await resolve_registration_owner_id(session, protocol.user_store, key)
 
 
 # Registration endpoints
--- a/core/switch_core/bridges/agent/api/handlers.py
+++ b/core/switch_core/bridges/agent/api/handlers.py
@@ -154,7 +154,10 @@ async def _resolve_registration_user_id(
     try:
         return await resolve_registration_owner_id(session, protocol.user_store, key)
     except RuntimeError as exc:
-        raise HTTPException(status_code=500, detail=str(exc)) from exc
+        logger.error("Agent-registration bootstrap owner resolution failed: %s", exc)
+        raise HTTPException(
+            status_code=503, detail="Agent registration is temporarily unavailable"
+        ) from exc
 
 
 # Registration endpoints
--- a/core/switch_core/bridges/agent/auth.py
+++ b/core/switch_core/bridges/agent/auth.py
@@ -11,6 +11,7 @@ from starlette.responses import Response
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from switch_core.bridges.agent.api_key_cache import ApiKeyCache
+from switch_core.bridges.agent.registration_bootstrap import REGISTRATION_KEY_TYPES
 from switch_core.db.models import Agent, ApiKey
 from switch_core.db.stores.agent_store import AgentStore
 from switch_core.db.stores.api_key_store import ApiKeyStore
@@ -71,8 +72,9 @@ class BearerAuthMiddleware:
        Used by agent-bridge HTTP endpoints and MCP.
     2. OIDC token — validated against the configured issuer, mapped to an
        agent via the ``oauth_client_id`` field on Agent.
-    3. Registration token — an ApiKey row of type ``"registration"`` that
-       has no associated agent. Used by the registration endpoint
+    3. Registration token — an ApiKey row of type ``"registration"`` or
+       ``"bootstrap"`` (see ``registration_bootstrap.py``) that has no
+       associated agent. Used by the registration endpoint
        (``POST /agents``). Sets ``scope["api_key"]`` so downstream
        handlers can validate it again.
 
@@ -137,7 +139,7 @@ class BearerAuthMiddleware:
         # Registration token: pass through (handler validates again). MCP rejects.
         if (
             api_key is not None
-            and api_key.type == "registration"
+            and api_key.type in REGISTRATION_KEY_TYPES
             and not path.startswith("/mcp")
         ):
             scope["api_key"] = api_key
--- a/core/switch_core/bridges/agent/protocol/service.py
+++ b/core/switch_core/bridges/agent/protocol/service.py
@@ -51,6 +51,10 @@ from switch_core.bridges.agent.protocol.types import (
     ToolCallReport,
     ToolSpec,
 )
+from switch_core.bridges.agent.registration_bootstrap import (
+    REGISTRATION_KEY_TYPES,
+    resolve_registration_owner_id,
+)
 from switch_core.bridges.resource.service import ResourceService
 from switch_core.crypto import encrypt_token
 from switch_core.db.models import (
@@ -435,14 +439,18 @@ class ProtocolService:
     ) -> RegistrationResult:
         """Resolve a registration token to its owner, then register the agent.
 
-        Raises PermissionError if the token does not match a stored ApiKey.
+        Raises PermissionError if the token does not match a stored ApiKey of
+        a registration type (see ``registration_bootstrap.REGISTRATION_KEY_TYPES``).
         Raises AgentExistsError if the name is taken and ``overwrite`` is False.
         """
         token_hash = hashlib.sha256(registration_token.encode()).hexdigest()
         async with self.session_factory() as session:
             key = await self.api_key_store.get_by_hash(session, token_hash)
-        if key is None:
-            raise PermissionError("Invalid registration token")
+            if key is None or key.type not in REGISTRATION_KEY_TYPES:
+                raise PermissionError("Invalid registration token")
+            owner_id = await resolve_registration_owner_id(
+                session, self.user_store, key
+            )
 
         return await self.register_agent(
             name=name,
@@ -453,7 +461,7 @@ class ProtocolService:
             tools=tools,
             models=models,
             metadata=metadata,
-            owner_id=key.user_id,
+            owner_id=owner_id,
             overwrite=overwrite,
             addressable_by_agent_ids=addressable_by_agent_ids,
             owner_only=owner_only,
--- a/core/switch_core/bridges/agent/protocol/service.py
+++ b/core/switch_core/bridges/agent/protocol/service.py
@@ -204,6 +204,17 @@ class AgentExistsError(Exception):
     caller did not opt into re-registration via ``overwrite=True``."""
 
 
+class BootstrapOwnerMisconfiguredError(Exception):
+    """Raised when the deployment-wide bootstrap key's owner account cannot
+    be resolved or trusted (see ``registration_bootstrap.py``).
+
+    Deliberately not a ``PermissionError``: the presented token is valid, so
+    this is a deployment misconfiguration, not a bad credential. A caller
+    that catches ``PermissionError`` to mean "bad token" must not also catch
+    this and report the same thing.
+    """
+
+
 def _describe_room(room: Room) -> RoomDescriptor:
     return RoomDescriptor(
         id=room.id,
@@ -440,10 +451,13 @@ class ProtocolService:
         """Resolve a registration token to its owner, then register the agent.
 
         Raises PermissionError if the token does not match a stored ApiKey of
-        a registration type (see ``registration_bootstrap.REGISTRATION_KEY_TYPES``),
-        or if the bootstrap owner cannot be resolved (misconfigured deployment
-        — the detail is logged, not raised, to match the HTTP registration
-        path's handling of the same failure).
+        a registration type (see ``registration_bootstrap.REGISTRATION_KEY_TYPES``).
+        Raises BootstrapOwnerMisconfiguredError if the token is valid but its
+        bootstrap owner cannot be resolved or trusted — a deployment fault,
+        not a bad credential, so callers must not conflate it with the
+        PermissionError above (the detail is logged here, not in the
+        exception, to match the HTTP registration path's handling of the
+        same failure).
         Raises AgentExistsError if the name is taken and ``overwrite`` is False.
         """
         token_hash = hashlib.sha256(registration_token.encode()).hexdigest()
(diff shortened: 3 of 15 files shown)


wojtyniakAQ:
Follow-up commit fixes a bug in the seeding logic found during self-review: the durable "already seeded" flag was only persisted when a brand-new key was created, not when an existing deployment's legacy key was migrated onto the new bootstrap type. That left a narrow window right after an upgrade where revoking the freshly migrated key and restarting before any other seed call had backfilled the flag would silently bring it back. Added a regression test (`test_revoking_a_migrated_legacy_key_survives_a_restart`) and an integration test that drives a bootstrap-registered agent through the real authorization path (`_require_room_action`) against a private room, rather than only asserting on ownership attributes.

wojtyniakAQ:
recheck

amaudruz (approved):

PR #400, opened 2026-09-08
docs: gateway OIDC sign-in setup and local development guide

wojtyniakAQ:
## Summary
- Document the gateway's bring-your-own OIDC browser sign-in path: the `GATEWAY_OIDC_*` variables, the all-or-nothing validation, the exact callback path, and what just-in-time provisioning does today (`docs/old/GATEWAY_OIDC_SETUP.md`).
- Cover WorkOS specifically, since it has two sign-in surfaces and only one (Connect, at the environment's AuthKit domain) is standards-compliant OIDC — the other looks like it works and doesn't.
- Add a local-development guide covering `just init-env`, `just up`, `just migrate`, `just run`, and `just gateway-dev` — including that `just up` does not start the gateway frontend and nothing warns you (`docs/old/LOCAL_DEVELOPMENT.md`).
- Add the `GATEWAY_OIDC_*` block to `.env.example`, commented out with placeholders.
- Link both new pages from `CLAUDE.md`, `CONTRIBUTING.md` and `docs/README.md`, following how existing `docs/old/` pages are indexed.

## Test plan
- [x] Verified every variable name and `just` recipe mentioned against `core/switch_core/config.py`, `core/switch_core/gateway/oidc_routes.py`, `core/switch_core/gateway/app.py`, `core/switch_core/db/stores/user_store.py`, the `justfile`, `deploy/local/docker-compose.yml`, `gateway/vite.config.ts` and `.env.example`.
- [x] No real credentials, internal hostnames or personal emails — only placeholders.
- [x] Only markdown and `.env.example` changed; no code touched, so no lint/typecheck run.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -82,7 +82,7 @@ MATTERMOST_USER_PASSWORD=
 # OAUTH_VERIFY_ISSUER=true
 
 # ── Gateway OIDC login (optional — bring your own IdP for the browser login;
-# distinct from the MCP OAuth_* block above) ─────────────────────────────────
+# distinct from the OAUTH_* block above) ──────────────────────────────────────
 # See docs/old/GATEWAY_OIDC_SETUP.md before setting these, especially if the
 # provider is WorkOS: it needs a Connect application, not a User Management
 # client id. Issuer/client id/secret are all-or-nothing — set all three or
--- a/.env.example
+++ b/.env.example
@@ -80,3 +80,18 @@ MATTERMOST_USER_PASSWORD=
 # OAUTH_ISSUER_URL=https://your-idp.example.com/realms/switch
 # OAUTH_AUDIENCE=switch
 # OAUTH_VERIFY_ISSUER=true
+
+# ── Gateway OIDC login (optional — bring your own IdP for the browser login;
+# distinct from the MCP OAuth_* block above) ─────────────────────────────────
+# See docs/old/GATEWAY_OIDC_SETUP.md before setting these, especially if the
+# provider is WorkOS: it needs a Connect application, not a User Management
+# client id. Issuer/client id/secret are all-or-nothing — set all three or
+# none. Quote GATEWAY_OIDC_SCOPES; it must include "openid" or no id_token is
+# issued.
+# GATEWAY_OIDC_ISSUER_URL=https://<your-idp-or-authkit-domain>
+# GATEWAY_OIDC_CLIENT_ID=<client-id>
+# GATEWAY_OIDC_CLIENT_SECRET=<client-secret>
+# GATEWAY_OIDC_SCOPES="openid profile email"
+# GATEWAY_OIDC_PROVIDER_LABEL=SSO
+# GATEWAY_OIDC_REDIRECT_URL=http://localhost:8000/gateway/auth/oidc/callback
+# GATEWAY_OIDC_REQUIRE_EMAIL_VERIFIED=true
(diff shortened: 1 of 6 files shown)


wojtyniakAQ:
Fixed the Switch Console local-dev advice: Gateway URL should be http://localhost:5173 (Vite dev server), API URL http://localhost:8000, with just gateway-dev running — not both pointed at :8000 as originally written.

Gateway URL is what Console opens directly for the "Full admin interface" button and what its reachability check pings, not just an alternate API base. Pointing it at :8000 makes the Open button load a bare API response (switch-core never serves the dashboard), while forgetting to start just gateway-dev makes a perfectly healthy backend show as an unreachable server — confirmed against console/apps/switch-console-desktop/src/main/core/switch-servers/{view.tsx,gateway-client.ts} (openGatewayPage uses server.gatewayUrl directly; gatewayFetch composes ${gatewayUrl}/gateway${path} for all of Console's own session-cookie calls) and onboard-location-agents.ts/add-agent.ts (server.apiUrl is what's written into each connected agent's own config as its endpoint).

wojtyniakAQ:
Fixed both review findings:

- docs/old/GATEWAY_OIDC_SETUP.md: the quoting-failure passage named the wrong tool. It's just's dotenv loader (set dotenv-load := true, every recipe including just run goes through it) that fails loudly on an unquoted, spaced GATEWAY_OIDC_SCOPES: "error: failed to load environment file ... Error parsing line", exit 1. uv run on its own doesn't read .env at all (silent, unset var); uv run --env-file .env only warns and continues. Rewrote the passage to name just correctly and to say plainly that running switch-core outside just loses that protection — the variable ends up silently unset, and the failure only surfaces later at the callback as a missing id_token, with nothing pointing back at the scope config.
- .env.example: fixed "the MCP OAuth_* block above" to "the OAUTH_* block above" to match the real variable names.

Left the application_not_found / User Management discovery / dashboard-only-credential claims as written, per the note that those were observed first-hand and aren't independently reviewable without provisioning a WorkOS environment.

wojtyniakAQ:
Fixed the remaining review finding: the claim that a missing openid scope surfaces at the callback "as a missing id_token" was wrong and has been dropped rather than replaced with another guess.

Traced the actual flow in oidc_routes.py: authorize_access_token doesn't raise on a missing id_token (authlib only populates token["userinfo"] when both an id_token and a nonce are present, otherwise it returns quietly). oidc_callback then falls through to a live client.userinfo(token=token) call against the discovered userinfo endpoint, and the outcome depends entirely on the provider — it may return claims and succeed anyway, or raise an unhandled httpx.HTTPStatusError (nothing catches it; the except OAuthError only wraps the earlier authorize_access_token call), giving the user an unhandled 500 rather than the 401 the page describes for the neighbouring "claims missing email/sub" case. Neither outcome is a recognizable "check your scopes" symptom, so the sentence now just says the request falls back to the provider's default scope with nothing pointing back at the config, without claiming a specific downstream failure.

Follow-up worth a ticket (not fixed here, and not touched — oidc_routes.py is being edited on another branch): the unhandled httpx error from that userinfo call is a real robustness gap in oidc_callback. An IdP hiccup at that point currently becomes an unhandled 500 instead of a clean 401/error response. Belongs in code, not in this docs PR.

wojtyniakAQ:

amaudruz (approved):

wojtyniakAQ:
Merge-order note.

This page documents `GATEWAY_OIDC_SCOPES` as optional, which is correct for `main` today. #403 makes it required and adds a render-time check in the Helm chart, so four passages here go stale the moment that lands.

Suggested order: this PR merges first, then #403 rebases and updates those four spots as part of the change that causes them — code and docs moving together. Noted on #403 as well so whichever lands second picks it up.

PR #407, opened 2026-09-08
Group templates: provision multiple rooms from one document (CHOO-2657)

abeldantas:
Right now one template file makes one room. We add a second document in the schema (a `group:` key instead of `room:`) so a single YAML file can declare a room group, several rooms inside it, and directed links between them.


abeldantas:
Folded into #409: the core half (group document, provision_group, aliases, create_room_from_yaml) is there as commit 46e4c4d3, merged onto main's parse_template and kickoff shape, plus the Console side (a Use page that creates agents and rooms from one document). The Console files in this diff were drift from an older base and are already on main.

PR #408, opened 2026-09-08
docs: hosted agents design spike

wojtyniakAQ:
Adds `docs/old/hosted-agents.md`, a design spike on what it would take for Switch to run an agent session **for** a customer, so that someone can sign up and have a working agent without installing anything. Today every agent runs on a machine the user owns. Marked as a spike and not implemented, in the style of `docs/old/multi-tenancy.md` (which is itself in flight on an unmerged branch, and is named as such in the text since this document references it throughout).

## The argument

**The runtime is mostly done and on the wrong side of the wire.** The Console sidecar is already a complete headless agent runner — around 2,400 lines, no Electron by construction, `tmux` plus `git` plus a Node floor of 18, with `AgentLaunchSpec` as a plain-JSON launch recipe and a two-method `SidecarHost` seam whose only implementation happens to be SSH. Put it in an image and the runtime problem is largely solved. What gets discarded is everything that assumes a person owns the box, most consequentially the assumption that a model credential is already on the machine because somebody typed it there.

**What has to be built is a control plane, a chat surface and a credential model.** Nothing in `switch_core` can start anything. The watcher that notices "this agent was addressed and has no session" lives inside the sidecar, which is circular for hosting; the server already models `auto_session` and already posts the "starting a session" message, so the model is there and only the actor is missing. And there is no surface a person can point a browser at: no message endpoints on the gateway API, no composer in the dashboard.

**Licensing decides the credential model, and it corrects an assumption elsewhere.** Anthropic's terms permit running Claude Code in hosted sandboxes on two conditions — an unmodified binary, and no paying for or intermediating usage on end users' behalf. So bring-your-own-credential is permitted for the paid path and a free tier on our Claude tokens is not. Beyond this document: the multi-tenancy spike's quota phase opens with "one customer's runaway loop is our bill", and against these terms that premise does not hold. The hard stop moves from tokens, which we cannot enforce because Switch is not in the request path, to compute seconds, which we own outright.

**A free tier is still possible, on a different vendor's model** — an open model served through our own account, with the customer connecting their own credential for the good models. That rests on a reading of the applicable model licence which is **explicitly marked unverified in the text** and listed as an open question: if the operative instrument restricts use to the subscribing customer's internal business purposes, that tier does not exist. Self-hosting the model is ruled out with numbers at twenty to a hundred and sixty times the API price for a free tier's duty cycle.

**ECS Fargate in our own account, and not on price.** Per-task isolation quoted from AWS, a network interface in a VPC we control so egress policy is a security group, a real IAM boundary, and residency that follows from the account. Cold start (thirty to forty-five seconds, billed from image pull) is named as the central risk — and the document now shows the case where the recommendation loses: warm Fargate is about $86 a month against roughly $5 for E2B and $9 for Sprites asleep, and they beat it below roughly seventy and forty per cent duty cycle respectively, while also waking in under a second. The recommendation is defended on subprocessor and account-control grounds and says in terms that it should not be defended on price.

**And an honest case against.** Hosting makes us a processor of customer code and prompts, which contradicts the README's own "designed to be self-hostable and for your data to stay where it is". §10 argues the objection rather than dismissing it, and offers the middle path: host the demo, not the product.

## Decisions the reader is being asked to make

The full table is in §12. The contested ones:

1. **ECS Fargate over the specialist sandbox vendors**, accepting that they are cheaper *and* faster to wake for a mostly-idle agent, and that the case is entirely about not adding a subprocessor and controlling the account. If phase 0's cold-start measurement forces a warm pool, the price argument inverts and this should be reopened.
2. **Bring-your-own-credential as the only model for the paid path.** Forced by the licence, and it puts an API key in the signup flow, which is where strangers leave.
3. **A free tier on an open model through our own account**, gated on verifying the licence claim, with the recommendation being a *validation process* against the real harness rather than a model name — every cheap model surveyed has a dated failure mode in this class of harness, and the inference backend alone can move a score by forty points.
4. **Hard-stop on compute seconds rather than tokens** on the paid path; on the free tier we own the account, so the provider-side cap returns.
5. **A minimal message list and composer in the gateway** as the first chat surface, ahead of the official Slack app — the decision most in tension with "Switch is not a messaging app", taken because signup cannot depend on a third party's install review.
6. **Hosted sessions do not inherit today's permission defaults**, and the mediation hook is not counted as a control at all.
7. **Sequencing.** Phases 0 to 3 are useful to current users and need nothing from multi-tenancy. Phase 5 is labelled a paid-only beta because it still contains the enrollment step phase 6 exists to remove; the two may want to ship together.

## Open questions, including two that can kill it

**Will we agree to Anthropic's Commercial Terms of Service, and who signs?** No agreement, no hosted Claude Code, and phase 1 does not exist. **Will the positioning change, and when?** §10 argues that shipping a hosted tier while the public material still says data stays where it is is the worst outcome available, and nobody has decided.

Then: what the free tier's model licence actually says; acceptable wake latency; how durable per-agent state is carried, given a Fargate task cannot attach a volume after launch (so a warm pool buys the infrastructure half of wake latency and nothing else, permanently); the idle-to-active ratio, measurable today; whether an inline proxy carrying the customer's own key counts as "intermediating"; how an end user completes an interactive sign-in without us intermediating it; what session resumption means once a destroyed task takes the CLI's context with it; per-tenant attribution for free-tier spend; where the single-replica constraint breaks; and who is on call.

Two market claims in §10 are labelled in the text as an unsourced read rather than a citation. Northflank is named as **not evaluated** and pushed into the phase 0 measurement rather than rejected on grounds nobody established.

## Scope

Two files: the document, and one sentence in `docs/README.md` so the description of `docs/old/` accounts for design spikes. No code, no version bumps, no changelog. Codebase claims were verified against the tree; the document has been through two review passes and the second one's corrections are in the third commit.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,7 +19,8 @@ site presents them.
 ## `old/`
 
 Design and operator material written for this repository: the architecture
-overview, the agent protocol, and per-bridge setup guides. None of it is
+overview, the agent protocol, per-bridge setup guides, and design spikes for
+work that has not been built. A spike says so in its first line. None of it is
 published, and none of it is covered by the pages under `official/` — where the
 two describe the same thing, this is the deeper account and the published page is
 the one users act on.
--- a/docs/old/hosted-agents.md
+++ b/docs/old/hosted-agents.md
@@ -19,6 +19,11 @@ agent with nothing installed. It is an argument rather than an agreed plan, and
 The deliverable is the design, the order of work, and the questions still open.
 No code was written.
 
+It refers throughout to the multi-tenancy design spike, which is
+`docs/old/multi-tenancy.md` — **in flight on an unmerged branch**, so it is not
+on `main` and a reader of this document may not be able to open it yet. Where
+this document quotes it, the quote is reproduced in full for that reason.
+
 ---
 
 ## 1. Where we are today
@@ -34,13 +39,14 @@ addressed to it, spawns the agent CLI under `tmux`, types prompts into the
 running terminal UI, and keeps a session registry on disk that it reconciles
 against live `tmux` panes when it restarts.
 
-It needs Node 20, `tmux` and `git`, and nothing else. The build script *enforces*
-that: a custom esbuild resolver fails the build if Electron, the local SQLite
-database or the ORM are ever pulled in transitively. The sidecar cannot acquire
-a desktop dependency by accident.
+It needs `tmux`, `git`, and Node 18 or newer — the preflight's floor, though the
+build targets Node 20, now past end of life, so a new image should pin 22 or 24.
+Nothing else, and the build script enforces it: a custom esbuild resolver fails
+the build if Electron, the local database or the ORM are pulled in transitively,
+so the sidecar cannot acquire a desktop dependency by accident.
 
-It also touches SSH nowhere. SSH is how Switch Console *delivers* it, and the
-seam between the two is two methods:
+It contains no SSH code — only comments mentioning it. SSH is how Switch Console
+*delivers* it, and the seam between the two is two methods:
 
 ```ts
 export interface SidecarHost {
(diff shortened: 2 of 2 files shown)


wojtyniakAQ:
Closing: hosted agent sessions are dropped as a direction for now, so this design should not sit open implying otherwise.

The branch stays if anyone wants the analysis later. The parts with a life beyond hosting, in case they are useful elsewhere: Anthropic permits running Claude Code in a hosted sandbox but forbids a platform paying for or intermediating an end user's model usage, which rules out a free tier on our own Claude credentials; AWS Fargate gives per-task microVM isolation but takes thirty to forty-five seconds to start a task; a Fargate task takes its volume configuration at launch and cannot attach an existing volume, so a warm pool cannot adopt a specific agent's state; and the gateway has no chat surface at all, since message read and write exist only on the agent-facing API.

Reopen rather than rewrite if the direction comes back.

PR #411, opened 2026-09-09
CHOO-2667 — the server side of the session interaction contract

simonflk-sb:
# CHOO-2667 — the server side of the session interaction contract

#394 builds the half that renders a session into Slack and turns a press back
into a `Command`. It stops at the shape of the routes and is driven from
fixtures, because the server half does not exist: nothing stores a session,
nothing owns its epoch, nothing decides which room may see it, and nothing
accepts a command.

This branch is that half — session state, leases, host ingest, room association,
publication policy, command acceptance and delivery. The rev 4 plan it follows
was moved out of #394's description into a room document so that PR describes
only the changes it carries.

**Not for merge.** Feature-branch work, one slice at a time.

### Base

Targets `work/rich-session-activity-slack` (#394), not `main`. The session
projection, the contract mirror and the renderers live there, and everything
here is built to be driven by them. Reviewing against that base means only the
new code is in the diff.

---

<details>
<summary><b>The plan this branch follows</b> — server-side session integration, rev 4</summary>

Reproduced from the room document so it is readable from here. The room
document is the one that gets revised; this is a copy as of the last update
to this description.

## Server-side session integration — proposed plan (rev 4)

> **Status: proposal, not started.** Revised 2026-09-09. The assumptions section
> below records provisional defaults so the plan does not block on anyone's
> approval; they are working decisions, not gates.

The slices above render and answer a session that does not exist. This section
covers the missing half: receive a real host's events in core, retain them,
publish through the existing collaboration bridge, and send answers back.

### Who owns what

The SDK host has its own owner, on the other side of this boundary. It is
further along than "not started":
`origin/worktree-sdk-providers@2469a180` has a persistent local host for all
five providers, with a durable command inbox and a replayable journal, and its
README names the boundary — local-only sessions, Switch credentials rejected,
"shared commands require the Switch server's lease, authorization and
reservation path". That path is ours.

### What already exists on our side

| Piece | Where | State |
|---|---|---|
| Contract types | `session/contract.py` | Complete, parity-tested against the TS examples |
| The fold | `session/projection.py` | Complete. Epoch fencing, revision dedupe, gap tolerant. Held across events; nothing long-lived owns one |
| Transport port | `session/transport.py` | `SessionEventSource` is two methods. Only `FixtureEventSource` implements it |
| Slack rendering | `session/renderers/slack.py` | Approvals, forms, turn activity, text fallbacks, budgets |
| Card persistence | `session_request_posts` + store | Complete. Token, handle, form, revision, four uniqueness constraints |
| Answer resolution | `session/inbound.py`, `form.py`, `text.py` | Complete. Press and typed answers both produce a contract-valid `Command` |
| Card redraw | `SessionRequestCards.refresh` | Written, no caller |

Reusable outside the session layer, and reused rather than reinvented below:

- Per-agent bearer auth (`BearerAuthMiddleware`, `get_agent_from_scope`) and the
  gateway's cookie auth at `/gateway`.
- The whole agent connection lifecycle: `ConnectionRegistry.open`/`beat`/`sweep`,
  generation-based takeover, room claim, `covers`.
- The SSE stream: `_frame` takes the event name from the event's own type, so a
  new event type costs no transport code; `Last-Event-ID` → `_resolve_start_cursor`
  → `Connection.cursor`, with explicit `gap` frames for all three loss modes.
- Media: `MediaBlob` / `MediaStore`, and `switch-media://<uuid>` handles minted
  at `transport/postgres.py:496`.
- Two precedents for exclusive ownership: `ConnectionRegistry.claim_room` and
  `role_leases` with heartbeat auto-release.
- The durable-record-plus-notification pattern: `notify_ddl.py` +
  `messages/notify.py` + `delivery_cursor_store.py`, where a lost notification
  costs latency and not data.

### What is genuinely missing

**Nothing about a session is persisted in core.** No sessions table, no host
session id, no event log, no command store, no lease, no room association.
`session_request_posts.session_id` is free text with no FK and nothing behind it.

**There is no authorization on a bridge answer, only identity.**
`_identify_actor` auto-creates a Matrix puppet for any external user in a
bridged channel. Whoever can see the channel can press the button. The contract
is explicit that this is not enough: "Permission to read a room or message an
agent does not imply permission to approve an action."

**Commands have nowhere durable to live.** `EventBuffer` is in memory,
single-process by construction (the Helm chart hard-fails on `replicaCount != 1`),
2000 events / 15 minutes, and its sequence resets to 1 on restart. It is a fine
notification channel and a non-starter as the record.

**Attachments — correction to rev 1.** Rev 1 said "no attachment service". That
was wrong. Core has one blob store (`MediaBlob`, the only `LargeBinary` column in
the tree), one minting site, and one opaque handle, and the upload route already
returns that handle to its caller. `Attachment.attachmentId` maps onto it with
nothing new to build. What is actually missing is narrower: every upload today is
also a room post (`send_media` requires room membership and posts an event), the
only authorization fact available is room membership (`media_blobs` has no owner,
room or session column), there is no user-authenticated read anywhere — the
gateway app registers no media router — and no renderer in `session/` reads
`Item.attachments` at all.

### The room association

Server-owned, and separate from the lease.

- **Lease acquisition and renewal do not authorise publication.** The lease is
  execution ownership: who may emit events for this session and under which
  epoch. It is written by the host with an agent token. Publication authority is
  a different fact with a different writer, so it is a different record —
  `session_room_associations`, not a column the lease path can touch.
- **Derived where the server can derive it.** When a session is started by an
  in-room command, the server already minted that message's `origin` from a real
  room message it verified. That origin is the association: room, and the actor
  who asked. Nothing the host says is consulted.
- **Explicit and authorised otherwise.** Any other association is a deliberate
  call by an actor the server has authorised in that room. A host cannot
  associate itself.
- **No association means no publication.** A Console-started session publishes
  nothing anywhere, ever, until someone with the right to do so associates it.
  That is a first-class acceptance scenario, not a degenerate case.

This is the contract's own §8.4 conclusion — "Room publication requires an
explicit association from session orchestration".

### Publication policy

With `audience` gone, what reaches a channel is a policy the server states, not
a field it reads and not "everything from an associated session". One function,
`publication_policy(...) -> decision`, tested on its own so the rule lives in one
place rather than spread through renderers.

**Transcript text is never published.** Not assistant text, not a final
response, not a user message typed in the Console. A room reply is something the
agent chooses to say, through the existing MCP `post_message` path, exactly as it
does today. The session integration does not become a second way for an agent to
talk to a room, and there is nothing to deduplicate because there is nothing
competing.

What "final state" means here is therefore narrow: activity status reaching its
terminal value, and request outcomes being shown. Not a transcript.

Consequence worth stating: `render_activity` as landed in slice 6 renders message
sections *and* the disclosure. Those message sections belong to the Console read
path, not the room path. S5 below gates on the policy rather than calling
`render_activity` wholesale.

#### Which requests may be published

1. Only from a session with a room association, and only to that room's channel.
2. Every open request of an associated session is published. Withholding a
   request from the room the session is associated with only means nobody there
   can answer it, and the Console still can.
3. A request that was already settled before it could be published is not posted
   as a live card. Its outcome is posted instead.
4. Publishing a card grants nothing. Authorization is decided per actor at answer
   time, in the acceptance service. A card visible to a channel is not permission
   for everyone in that channel.
5. A card is updated in place through `submitting`, `resolved` and `closed`.
   Where an edit fails or the platform cannot edit, a status reply is posted
   instead (§8.3).

#### Which activity summaries may be published

1. Activity is never a post of its own. It appears only as the collapsed
   disclosure attached to the turn's existing post, and as edits to it.
2. What is published is the item's `title` and `status`, plus its `text` where
   the host has written a short summary — "4 files", "1 failed, 41 passed". The
   contract already obliges the host to exclude raw provider payloads from shared
   events; we bound the result anyway with the budgets slice 6 built.
3. Frequent revisions are combined before publication, not posted one per
   revision (§8.4).
4. An item whose `origin.messageId` names a message already in that room is never
   republished. It is used to anchor and thread.
5. Everything not published here remains readable through the gateway read
   routes, under their own authorization.

### Command acceptance, including reservation

**One service, two entrypoints.** `SessionCommandService.accept(actor, session_id, body)`
implements the contract's §5.1 steps in order — identity, session and room
access, operation permission, epoch and capability, destination and size,
`(sessionId, commandId)` idempotency, **atomic reservation for an answer**,
durable save, 202, dispatch, `dispatched`. Nothing about acceptance lives in a
caller.

- **Gateway entrypoint**: `POST /gateway/v1/sessions/{id}/commands`, cookie auth,
  actor is the signed-in Switch user.
- **Bridge entrypoint**: in-process, not HTTP. The bridge has already verified the
  platform callback with the platform's own method and mapped it to a Switch
  actor; it hands that verified actor to the service and nothing else.
  `_drop_session_command` becomes this call.

The service authorises. The bridge supplies identity, and identity is all it has:
`_identify_actor` will hand it a puppet for a stranger in a bridged channel.
Authorization for a session command is therefore a check the service performs on
the Switch actor, and it is not satisfied by channel membership.

**Reservation happens inside acceptance, before anything is delivered.** The key
is the contract's `(sessionId, epoch, requestId, expectedRevision)`, taken
atomically as a unique constraint rather than in application logic — the same
shape as the four constraints already on `session_request_posts`. A command that
cannot take the reservation is rejected with `REQUEST_BUSY` and never reaches the
host. `request.submitting` is published after the reservation succeeds. This is
what makes two people pressing at once safe, and it cannot be a later slice
without the delivery path being wrong in between.

### Command storage, delivery and cursors

**Storage is the record.** `session_commands`, one row per command id, carrying
status through `accepted → dispatched → applied | rejected | unknown`. Everything
about replay, idempotency and "did this actually happen" is answered from the
table. Nothing about correctness depends on a process staying up.

**Delivery is replayable, not exactly-once.** No component promises a command is
delivered precisely once. What is promised is narrower and achievable: the
delivery cursor is durable, a command id is stable across every retry, and the
host prevents duplicate *execution* under a repeated id. Server acceptance cannot
guarantee one native SDK action after a process failure, and the contract says so.

- The route of record is `GET /agent/v1/sessions/{id}/commands?after=`, served
  from the table. Replayable, works for a session with no room, survives every
  restart. This is the floor.
- Low latency comes from a nudge on the SSE stream the host's agent already
  holds: a new `AgentEvent` type carrying no body and no ordering meaning,
  enqueued through `EventBuffer.enqueue` exactly as `command` is today. It rides
  the existing auth, connection lifecycle, `id:` space, cursor and gap frames.
- Two known extensions this needs: the `EventType` literal and payload map, and a
  deliberate branch in `is_notifiable`, whose fallback is
  `type.startswith("task_")` — a new type is invisible to every `filter=addressed`
  connection until it is added there. Plus a protocol version bump.

Routing caveat, stated rather than papered over: SSE routing is agent + room, so
the nudge only reaches a room-associated session. A session with no association
has no nudge and polls the durable route. That is acceptable precisely because
such a session publishes nothing and nobody in a channel is waiting.

#### Four cursors, deliberately separate

| Cursor | Scope | Assigned by | Used for | Recovery |
|---|---|---|---|---|
| `hostSequence` | Per session, per epoch | Host | Ordering and gap detection on ingest | A hole returns `EXPECTED_SEQUENCE`; the host retries from its outbox |
| Server `sequence` | Per session | Server | The gateway read stream; interleaves host events with server command and connection events | `CURSOR_EXPIRED` with `snapshotRequired`, then snapshot at N and subscribe after N |
| Command delivery position | Per session | Server, on durable insert | `commands?after=` | Durable. Unaffected by restarts and by anything the SSE stream does |
| Agent SSE `id:` | Per agent | `EventBuffer` | The nudge, and every other agent event | Existing `gap` frames; resets to 1 on restart, which costs nothing because the nudge carries no state |

They are not merged and none is derived from another. A second cursor *on the SSE
stream* was considered and rejected: `Last-Event-ID` is one scalar,
`Connection.cursor` is one int, and `ConnectionBeatRequest.cursor` is one int
clamped against one head, so a second sequence there would be a protocol break in
three files. The nudge therefore interleaves into the existing per-agent sequence
and carries no session ordering of its own.

#### Recovery after a lost notification

- **Nudge lost** (dropped frame, agent offline, `gap` frame, restart of the
  in-memory buffer): nothing is lost, because the nudge carries no state. The next
  nudge or the next poll of `commands?after=` delivers the command, and the
  delivery position is the only truth.
- **Host offline when a command is accepted**: for v1 the contract says reject new
  commands while the host is offline, and keep the status of commands already
  accepted. An accepted command survives as a row and is delivered when the host
  returns.
- **Core restarts between `accepted` and `dispatched`**: the row is `accepted`, so
  the command is delivered on the next read. Status is never silently lost.

#### Recovery after a failed or interrupted Slack publication

Correcting rev 3, which claimed the existing unique constraints made publication
idempotent. **They do not.** `(bridge_id, external_post_id)` can only fire once
the external post id is known, and the case that matters is precisely the one
where it is not: Slack accepted the post and we crashed, or timed out, before
recording it. The retry posts again, Slack mints a *different* id, and the
constraint is satisfied by two cards for one request. A database constraint
prevents two concurrent attempts. It cannot prevent a second sequential attempt
after an uncertain outcome.

What actually prevents it is an intent record written **before** the external
call, plus reconciliation against the platform before any re-post:

- `session_publications`, one row per `(bridge_id, session_id, request_id)`
  intent, moving `intended → in-flight → posted | failed`. The row is committed
  before the Slack call, so a crash leaves a durable `in-flight` marker rather
  than silence. The uniqueness constraint on that key is what stops two workers
  or two triggers racing; it is not what stops a retry.
- Every card already carries an opaque token (`session_request_posts.token`,
  unique, embedded in the card's action values). That token is the reconciliation
  key. On finding an `in-flight` row, the retry reads recent channel history for
  a post carrying that token, and adopts it if present. Only an inconclusive read
  leads to a new post.
- Slack has no idempotency key on `chat.postMessage`, so reconciliation is not a
  refinement — it is the only sound mechanism available.
- If reconciliation itself cannot answer (history unreadable, lookback exhausted),
  **we do not post blind.** The row stays `in-flight`, the failure is logged at
  error, and the request remains answerable in the Console. A missing card that is
  visibly missing is better than two cards for one decision, and the second card
  is the failure mode that puts two people on the same question.
- A failed **edit** falls back to a status reply in the same thread (§8.3), so a
  settled request is never left showing an open card. If the reply also fails, the
  stored revision is not advanced and the intent stays unresolved for the sweep.

#### Retries do not wait for a session event

Also correcting rev 3, which retried "on the next projection change". That makes
recovery depend on the session producing more events, so a request whose session
goes quiet — which is exactly what a session waiting on an approval does — is
never retried at all.

Retry is a background sweep over unresolved `session_publications` rows, on its
own timer with backoff, independent of the event stream. `main.py` already runs
sweep loops of this shape for connections and runtime state; this is another. A
session that emits nothing further still gets its card.

### Slices

Each ends in something observable. S1–S5 need nothing from the host side: the ingest route
can be driven by replaying the recorded fixture's host events, which retires
`!session-demo` in favour of a path where only the originating process is a
stand-in.

| # | Slice | Ends in |
|---|---|---|
| S1 | Session state at rest — `sessions`, `session_events`, `session_commands`, `session_leases`, `session_room_associations`, `session_publications`, stores, migration | Store tests against real Postgres, including the reservation constraint |
| S2 | Lease — create/renew, server assigns the epoch, second holder refused, and no path from the lease to a room | An agent token can take and lose a lease; a leaseholder still cannot publish |
| S3 | Host ingest — `POST /agent/v1/sessions/{id}/events`, validated, epoch and `hostSequence` checked, server `sequence` assigned | Replaying the fixture over HTTP builds the same projection the fixture source does |
| S4 | Association and publication policy — derive from a verified room command, explicit authorised association otherwise, `publication_policy` as its own tested unit | An unassociated session provably posts nothing; an associated one posts requests and activity and no transcript text |
| S5 | The publisher — a per-session `SessionProjection` held live behind a DB-backed `SessionEventSource`, driving cards and the activity disclosure through the policy; intent written before the Slack call, token reconciliation before any re-post, and a sweep that retries on its own timer | Real events post real cards in Slack; a forced edit failure still shows the outcome; a kill between the Slack call and the write produces one card, not two, with no further session events |
| S6 | Command acceptance — §5.1 in order including atomic reservation, gateway route and the bridge's trusted in-process entrypoint, authorization, idempotency, durable save | An unauthorised press is refused; two competing presses produce one reservation and one `REQUEST_BUSY`, before anything is delivered |
| S7 | Delivery — durable `commands?after=` route plus the SSE nudge, ack through the existing ops door | A Slack press reaches a host process and survives a restart, with a stable command id on every retry |
| S8 | Settlement — `request.settled` ingest, `command.status`, card refresh to `resolved`/`closed` showing the deciding actor and surface | Press in Slack → host acts → the card updates itself |
| S9 | Gateway read routes — snapshot, `events?after=`, command submit and status | The Console's existing `SessionTransport` points at Switch instead of loopback |

Slices 7 and 8 from the list above fold in: end-of-turn becomes natural once the
publisher holds a per-turn anchor, and the parked revert is independent of all of
this.

### Current implementation assumptions

Working decisions taken so the plan can proceed. Each is provisional, each is
revisable, and none is an approval gate.

**Session registration.** The host generates `sessionId`. The first authorised
lease acquisition registers it against the authenticated agent. An ownership
conflict is rejected, and event ingestion for an unregistered session is
rejected.

**Epoch and recovery.** Events from a stale epoch are rejected. History already
accepted is preserved as it was: old commands are not relabelled and uncertain
commands are not automatically replayed. What the host does with its own journal
under a superseded epoch is the host's business and is not specified here. Epoch
recovery is recorded as an integration checkpoint to work through with the host
rather than something either side settles alone.

**Attachments.** Session attachment transfer is deferred for this
implementation. `Attachment.attachmentId` is specified now as a
`switch-media://` handle so nothing has to be renamed later. Where a command or
item carries an attachment, the unsupported operation is reported explicitly —
never dropped silently — and the usable text is preserved and published rather
than the whole item being refused. How the Console displays a deferred
attachment is the Console's decision and is not prescribed here. Undeferring
costs roughly one migration, one store method, two routes and one authorization
predicate on top of the media store that already exists, so it is a slice rather
than a redesign if it turns out to be needed inside v1.

### Acceptance scenarios

Beyond the contract's §10 list:

- **Console-started session, no association.** Events ingest, the gateway shows
  everything, no channel anywhere receives a post — including request cards.
- **Association cannot be self-granted.** A host holding a valid lease, renewing
  it, cannot cause publication. A `session.upsert` naming a room does not
  associate.
- **No transcript text in a channel.** An associated session runs a full turn with
  assistant output. The channel shows the activity disclosure and any request, and
  no assistant text. What the room hears is what the agent chose to say through
  `post_message`.
- **Unauthorised answer.** An actor who can read the channel and has no approval
  permission presses the button: `NOT_AUTHORIZED`, the card is unchanged, no row
  in `session_commands`, no reservation taken, and the request stays `open`.
- **Competing answers.** Two authorised users press different options within the
  same revision. One takes the reservation; the other gets `REQUEST_BUSY` before
  anything is delivered. Both cards end showing the same settled result and the
  same deciding actor.
- **Lost nudge.** The nudge is dropped and the SSE stream reports a `gap`. The
  command is still delivered from `commands?after=`, and re-delivery under the
  same command id does not produce a second turn.
- **Restart mid-flight.** Core restarts between `accepted` and `dispatched`. The
  command is still delivered and its status is never silently lost to `unknown`.
- **Interrupted publication.** The process is killed between Slack accepting the
  post and the row being written. On restart the retry finds the `in-flight`
  intent, reconciles on the embedded token, adopts the existing post, and one card
  exists. Forcing the reconciliation read to fail instead leaves the row
  unresolved and logs at error — and still does not post a second card.
- **Retry without a further session event.** The same kill, on a session that
  emits nothing else afterwards because it is waiting on the approval. The sweep
  publishes the card anyway.
- **Interrupted edit.** A card edit fails on settlement: a status reply appears in
  the thread, no request is left showing an open card, and the stored revision is
  not advanced.
- **An attachment nobody can carry.** An item arrives with an attachment. The text
  publishes, the unsupported operation is reported explicitly, and nothing is
  silently dropped.

</details>

---

## Slices

| # | Slice | Ends in |
|---|---|---|
| S1 | Session state at rest — the six tables, their stores and the migration | Store tests against real Postgres, including the reservation constraint |
| S2 | Lease — create/renew, server assigns the epoch, second holder refused, and no path from the lease to a room | An agent token can take and lose a lease; a leaseholder still cannot publish |
| S3 | Host ingest — validated events, epoch and `hostSequence` checked | Replaying the fixture over HTTP builds the projection the fixture source does |
| S4 | Association and publication policy | An unassociated session provably posts nothing |
| S5 | The publisher — live projection, intent before the call, reconciliation, sweep | Real events post real cards; a kill mid-post produces one card, not two |
| S6 | Command acceptance — §5.1 in order, including the atomic reservation | Two competing presses produce one reservation and one `REQUEST_BUSY` |
| S7 | Delivery — durable `commands?after=` plus the SSE nudge | A press reaches a host process and survives a restart |
| S8 | Settlement — `request.settled`, `command.status`, card refresh | Press in Slack → host acts → the card updates itself |
| S9 | Gateway read routes — snapshot, `events?after=`, submit and status | The Console's `SessionTransport` points at Switch instead of loopback |

- [x] **S1** — Session state at rest
- [x] **S2** — Lease
- [x] **S3** — Host ingest
- [x] **S4** — Association and publication policy
- [ ] **S5** — The publisher
- [ ] **S6** — Command acceptance
- [ ] **S7** — Delivery
- [ ] **S8** — Settlement
- [ ] **S9** — Gateway read routes

---

<details>
<summary><b>S1 — Session state at rest</b> ✅ landed</summary>

Six tables, one migration, six stores. Nothing is wired to a route yet: this is
the floor the rest stands on.

| | |
|---|---|
| `sessions` | One row per session id, the anchor everything hangs off |
| `session_leases` | Who owns execution, and under which epoch |
| `session_events` | The log — host events and server events in one order |
| `session_commands` | Commands the server accepted, and how far they got |
| `session_room_associations` | Which room a session may publish into, and who said so |
| `session_publications` | The intent to post, written before the platform is called |

**Worth calling out**

- **The reservation is a row, not a table.** An answer carries the request and
  the revision it answers, and a partial unique index over
  `(session, epoch, request_id, expected_revision)` on `session_commands` is
  what §5.2 asks be taken atomically. The insert either reserves and saves or
  fails having delivered nothing — there is no window where a reservation
  outlives the command that holds it, and nothing to clean up. It is partial
  because a prompt or an interrupt answers nothing and reserves nothing.
- **Positions are allocated under a per-session advisory lock, not by a
  sequence**, for the reason `MessageStore._next_seq` already sets out: a
  sequence numbers a row when its INSERT runs, not when it commits, so a reader
  paging on `> n` can step over a row that is still in flight and never come
  back for it. Both `session_events.sequence` and
  `session_commands.delivery_position` have exactly that reader. Two tests hold
  a pair of transactions open across a barrier and fail without the lock.
- **No `connectivity` column.** The server derives it from the lease and its
  heartbeat. A host's own claim that it is reachable cannot be the thing that
  answers whether it is. `session_events.epoch` is nullable for the same reason:
  the event that says a host has gone is emitted by the server exactly when
  there is no epoch left to name. An event that *does* carry a `host_sequence`
  must name its epoch, under a CHECK — the uniqueness index keys on the epoch
  and Postgres NULLs are distinct, so a host position under no epoch would be a
  position the index cannot refuse.
- **A lease is renewed and released by name.** Both take the `host_id` and
  `epoch` of the holder and raise when they do not match, so a host displaced by
  a newer epoch cannot keep its successor's lease alive or drop it.
- **The stores tell their constraints apart.** A repeated command id and a taken
  reservation are different answers in the contract — idempotency against
  `REQUEST_BUSY` — so they are raised as different exceptions rather than one
  `IntegrityError` whose meaning depends on which index Postgres checked first.
  Same for a replayed event id against a miscounted `hostSequence`. The inserts
  run in a savepoint, so the caller's transaction is still usable when it has to
  answer — and, more to the point, so a batch that holds one duplicate can
  commit the rest. A retry that fixes a gap must not be the thing that widens
  it.
- **The lease cannot reach a room.** Publication authority is a separate table
  with separate writers, so a host holding a lease with its own token still
  cannot publish itself into a room.
- **A publication intent is committed before the Slack call.** A card Slack
  accepted and we failed to record is indistinguishable from one never posted,
  and the uniqueness constraint cannot help because it keys on the external post
  id, which is exactly what is missing. `attempts` counts tries rather than
  failures, because a try that ended in silence is the one worth backing off on.
  Recording a post and recording a failure are separate calls, so a failed edit
  after a successful post cannot erase the one fact saying a card exists. The
  sweep claims its rows with `FOR UPDATE SKIP LOCKED`: reconciling takes a call
  to the platform, which can outlast the timer, and two sweeps adopting one
  in-flight intent post the second card this table exists to prevent.
- The ORM class for `sessions` is `HostSession`. `Session` is taken twice over —
  by `AsyncSession` and by the contract's own `Session` — and `agent_sessions`
  is a different thing entirely (an agent's transport slot in one room).

**Deliberately not done**

`session_request_posts.session_id` still has no foreign key to `sessions.id`.
Adding one now breaks the existing tests, which use a free-text session id, and
nothing on this branch writes that table yet. It belongs with S5, where the
publisher becomes the writer.

`host_id` lives on the lease and not on `sessions`, so a stopped session does
not remember which host last ran it. The contract wants it, and S9's snapshot
will need it — but S2 is the slice that writes the lease, so the column lands
there rather than sitting unmaintained for two slices. The migration is
unmerged, so that is an edit to the same file and not a second migration.

`session_leases.agent_id` and `session_room_associations.room_id` are unindexed.
`ix_sessions_agent` is there because "this agent's sessions" is a read both the
Console and the gateway make and `agents` cascades into it; the other two are
`CREATE INDEX CONCURRENTLY` material if a hot path turns up.

**Left for S5**

`claim_unresolved` holds its lock across the platform call, which makes the
sweep a fourth place a transaction stays open across a network round trip,
alongside the three `RoomService` methods named on
`db_idle_in_transaction_session_timeout`. The publisher decides which way to
settle it — keep the claim and add the sweep to that list, or hold the lock only
long enough to mark the row in-flight. Recorded on the method.

**Proof**

45 store tests against real Postgres. The three that matter most:

- a second answer to the same `(session, epoch, request, revision)` raises,
  while any number of non-answers queue freely;
- two overlapping transactions get two positions;
- a reader never sees a position above one that has not committed — one writer
  holds an allocated position open while another runs past it.

The last is the one that proves the allocator. It fails both with the advisory
lock removed and with the lock swapped for a real `nextval` sequence, which is
the regression the docstring argues against and which the first two tests happily
passed.

`test_migration_parity` already replays the chain into an empty database and
compares it against the models, so the migration is covered without a test of
its own — with one gap worth naming: alembic has no comparator for CHECK
constraints, so the two here (`ck_session_commands_answer_is_whole` and
`ck_session_events_host_event_names_its_epoch`) match the models by inspection
rather than by that test. They are the two things in the migration carrying a
rule rather than tidiness, so that is the narrowest part of the coverage.

`downgrade` was run by hand and leaves the schema as it was.

</details>

<details>
<summary><b>S2 — Lease</b> ✅ landed</summary>

One route: `POST /agent/v1/sessions/{id}/lease`. It is how a host gets into a
position to speak the contract at all, and it is the first thing on this branch
a token can reach.

```
POST /agent/v1/sessions/s-1/lease   {"hostId": "host-a"}
  → 200 {"sessionId": "s-1", "epoch": "9f3…", "hostId": "host-a", "displaced": null}

POST /agent/v1/sessions/s-1/lease   {"hostId": "host-a", "epoch": "9f3…"}
  → 200  (heartbeat, same epoch)

POST /agent/v1/sessions/s-1/lease   {"hostId": "host-b"}
  → 409 {"code": "LEASE_HELD", "message": "…", "retryable": true}

POST /agent/v1/sessions/s-1/lease   {"hostId": "host-b", "takeover": true}
  → 200 {"epoch": "c17…", "hostId": "host-b", "displaced": "host-a"}
```

**Worth calling out**

- **`/agent/v1` is new to this app.** Everything else on the agent bridge is
  unversioned under `/agents`. The contract names its own version segment and
  the host SDK builds its URLs from it, so these routes carry it. It is not in
  `PUBLIC_PATH_PREFIXES`, so the bearer middleware authenticates it by default
  and the handler is handed a resolved agent rather than a claim.
- **One route for create and renew**, as the contract has it. A body with no
  `epoch` is an acquisition and the server mints the generation; a body with one
  is a heartbeat under the generation the host already holds. A host cannot name
  its own epoch — if it could, a displaced host could name the one it just lost,
  and the fence the server puts in front of it would be a fence it holds the key
  to.
- **A live holder is refused; a stale one is free.** `takeover: true` is
  required to displace a host that is still heartbeating, mirroring
  `ConnectionRegistry.claim_room` — the usual cause of a collision is a stale
  process and refusing surfaces it. A lease whose `last_seen_at` has aged past
  the TTL is logically free without the flag, computed at read time the way
  `role_leases` already does it, so no reaper has to run for a dead host to give
  up its session. Requiring the flag on the ordinary restart path is exactly
  where it would stop meaning anything.
- **90s TTL**, against `role_leases`' 6s. The holder is a host process doing
  real work rather than a channel loop: at a 30s renew cadence this tolerates
  two fully missed renews, and the cost of being wrong is displacing a host that
  is merely slow.
- **The incumbent is not told.** It is displaced by the epoch changing under it
  and finds out when its next renew is refused. That is the only mechanism that
  also works for a holder that has stopped listening.
- **`takeover` does not cross the agent boundary.** A session belongs to the
  agent that registered it, and a session id is host-generated and not a secret,
  so a foreign agent gets `NOT_AUTHORIZED` before liveness is even considered.
- **The lease registers the session.** The contract has no "create session"
  call, and this is the first moment the server has an authenticated agent to
  bind the id to. `host_id` now lives on `sessions` as well as the lease —
  deferred from S1, folded into the same unmerged migration rather than a second
  one — because the lease row is deleted on release and a stopped session still
  has to say which host ran it.
- **A second error envelope, deliberately.** The session routes answer
  `{code, message, retryable}` per §6.2; every other route on the app keeps
  answering `{"detail": …}`. A status code alone cannot separate the contract's
  four different 409s, and the host branches on the code. It is applied at the
  door, not only by the handler: a rejected token and a body that failed to
  parse are the two failures a host is most likely to hit, and both are decided
  before the handler runs.
- **Timestamp columns on the `session_*` tables are now `Mapped[datetime]`**,
  unlike the rest of `models.py`, which annotates them `Mapped[str]`. Lease
  liveness is arithmetic on one, and typing it as text makes that comparison an
  error mypy is right about. Confined to the tables this branch introduces; the
  rest of the file is a separate change.

**Flagged for a decision**

Two codes are **not** in the contract's §6.2 table.

`LEASE_HELD` (409). The contract says the server permits one lease owner and
blocks the previous one, but gives no code for refusing the second claimant
while the first is alive, and none of the codes it does give is honest about it
— `NOT_AUTHORIZED` would tell a correctly configured host it may not run its own
session, which would send it away for good.

`INVALID_REQUEST` (422). The table's codes are semantic outcomes; none covers a
body that failed to parse. Without it, the one failure a host hits while it is
still being written is the one failure its decoder cannot read.

Both are one-line changes in `errors.py` and the skills if the contract should
say something else.

**Proof**

26 route tests, driven through the real router against real Postgres with only
the authenticated agent substituted. The three that matter most:

- a second live host is refused and the incumbent's row is untouched — both fail
  if the liveness check is removed;
- a takeover reports the host it displaced, which caught a real bug: the store
  updates the row the caller's object is mapped to, so reading `displaced` after
  the swap named the winner;
- a renewal under a displaced epoch is refused, which is the only way a
  displaced host learns it lost.

The acceptance criterion — a leaseholder still cannot publish — is pinned by a
test that drives every lease path (acquire, takeover, renew, second session) and
asserts `session_room_associations` is still empty. It is a guard rather than a
regression test: nothing writes that table yet, so it would pass on an empty
implementation too. It is here to fail the first time a lease path grows a
shortcut into a room.

**Review round 1** — `175d3fa7`

- **`/agent/v1` reached nothing in a real deployment.** Ingress `Prefix`
  matching is per path segment, so the existing `/agents` entry does not cover
  `/agent/v1/...`; a lease claim went to the gateway pod, whose nginx proxies
  `location /gateway` and nothing else. Added `/agent` to `agentApiPaths` and to
  the sample ingress. Not renamed to `/agents/v1`: §6 of the contract writes the
  path out as `POST /agent/v1/sessions/{id}/lease`.
- **A restarted host could not reclaim its own session.** It loses its epoch
  with its memory, so it must acquire rather than renew, and its own lease row
  is still well inside the TTL — so it was told to retry with `takeover`, making
  the flag the ordinary restart path. The host already named on the lease is now
  let through. Two live processes sharing one `host_id` are indistinguishable to
  this table either way.
- **The envelope now covers the door.** A rotated key returned
  `{"detail": "Not authenticated"}` and a malformed body `{"detail": [...]}`,
  neither with a `code`. Both now answer `NOT_AUTHORIZED` / `INVALID_REQUEST` in
  the envelope, scoped to the contract's paths so existing clients and the MCP
  door are unaffected.
- **The registration race gave the wrong answer to a different agent.** It
  returned a retryable `LEASE_HELD`, sending a caller back for a refusal it
  could have had immediately. It now reads the winner and answers
  `NOT_AUTHORIZED, retryable: false` when the session belongs to someone else.
- **`epoch` + `takeover` together is refused.** It reads as "renew, or take it
  back if I lost it", and resolving it either way is silent: as a takeover it
  hands back an epoch the host did not ask to keep, as a renewal it drops the
  flag.

Seven new tests, all of which fail with the fix backed out.

**Review round 2** — `01756a47`

- **404 and 405 on a session path escaped the envelope.** The handler was
  registered against FastAPI's `HTTPException`; Starlette dispatches on the
  raised type's MRO and the router raises the *parent* class, so a route this
  Switch does not have yet answered `{"detail": "Not Found"}` with no code —
  version skew, reported as an unknown error. Registered against Starlette's
  class, which covers both.
- **The fallback code contradicted the table.** Everything that was not 401/403
  became `INVALID_REQUEST` while keeping its status, so a 404 would have come
  back as `INVALID_REQUEST` at 404 against a table that says 422. The code is
  now derived from the status: 404 is `NOT_FOUND`, which the contract already
  has, and `INVALID_REQUEST` is the fallback for a status the contract does not
  name (405).
- **`displaced` no longer names the host itself.** A restart inside the TTL took
  the session from nobody, and anything branching on `displaced is not None`
  would have seen a takeover on every ordinary restart.
- **`exc.headers` survives** (`233a6c8e`). Taking `HTTPException` off the
  built-in handler took its header pass-through with it, dropping the router's
  `Allow` from every 405 in the app — the older `/agents` routes included. No
  consumer reads it today; a `WWW-Authenticate` later would have gone the same
  way silently.

**Operators upgrading need to act, and Helm will not do it for them.**
`ingress.mode` defaults to `existing`, which renders no Ingress at all — so the
`agentApiPaths` entry above only takes effect under `mode: managed`. Everyone
else is hand-maintaining a manifest that `helm upgrade` does not touch, and
`/agent` will keep falling through to the gateway on their cluster until they
add the rule. Noted in the chart README next to the agent API section; it wants
a line in the release notes too.

</details>

---

<details>
<summary><b>S3 — Host ingest</b> ✅ landed</summary>

`POST /agent/v1/sessions/{id}/events`. A host sends a batch of the events it
has queued, the server decides whether they may become durable, and the answer
says how far the outbox may be truncated.

```
POST /agent/v1/sessions/s-1/events  {"events": [ …hostSequence 1, 2, 3… ]}
  → 200 {"sessionId": "s-1", "epoch": "9f3…", "acceptedThrough": 3, "sequence": 3}

POST /agent/v1/sessions/s-1/events  {"events": [ …hostSequence 2, 3, 4… ]}
  → 200 {"acceptedThrough": 4, …}          (2 and 3 already held; 4 appended)

POST /agent/v1/sessions/s-1/events  {"events": [ …hostSequence 6… ]}
  → 409 {"code": "EXPECTED_SEQUENCE", "message": "… Resend from 5.", "retryable": true}

POST /agent/v1/sessions/s-1/events  {"events": [ …epoch the host no longer holds… ]}
  → 409 {"code": "STALE_EPOCH", "retryable": false}
```

**Worth calling out**

- **A repeat is the retry path, not an error.** A host truncates its outbox on
  `acceptedThrough`, so a lost response means it resends what the server already
  has. Refusing that wholesale would leave the outbox stuck forever. Repeats are
  matched per event — same id, same body — and skipped; only the remainder has
  to be contiguous. Different content under a position that is already taken is
  the opposite case, and it is `IDEMPOTENCY_CONFLICT` rather than an update.
- **All or nothing.** Every rule is checked before anything is written and the
  route commits once, so a refusal part way through a batch leaves the log
  untouched and the host's outbox exactly where it was. Otherwise
  `acceptedThrough` would move somewhere the host was never told about.
- **The head is read inside the lock `append` takes.** Deciding "already have
  it" or "next one" outside it leaves a window where two batches both read one
  position as free. `SessionEventStore.lock` is public for that; advisory locks
  are re-entrant within a transaction, so `append` taking it again costs
  nothing.
- **Two counters, reported together and never comparable.** `acceptedThrough` is
  the host's numbering under one epoch; `sequence` is the server's own log
  position, which also numbers events the host never sent. A host that treated
  `sequence` as an outbox cursor would delete work it had no answer for.
- **Sending events renews the lease.** A host emitting is provably alive.
  Without this a busy host that never gets round to a heartbeat becomes
  displaceable at the 90s TTL mid-turn.
- **A `session.upsert` may not name another agent.** `parse_host_event` checks
  the session id and epoch the body restates, because those are in the envelope
  beside it. The agent id is not — only the server knows who authenticated — so
  the route checks it, and the body is refused rather than published as fact to
  every later reader of the log.
- **The 1 MiB batch cap lives in the ingest service, not `contract.py`.** Only a
  server receives a batch; the host's own parser has no batch to measure, so
  mirroring it beside the 64 KiB event cap would claim a parity that does not
  exist.
- **The ownership check moved to `sessions/ownership.py`.** Lease and ingest ask
  it the same way, and a second copy would be a second answer to "may this agent
  speak for this session".

**Acceptance.** `examples.activity.json` — 11 recorded events — is converted to
host events, posted over HTTP, and folded back out of Postgres. The projection
that builds is the same object the fixture source builds directly, and the
stored bodies are byte-identical to what the host sent. The projection alone
would not pin the second half: Pydantic reads snake_case as happily as
camelCase, so a log written in the wrong dialect rebuilds fine here and breaks
the moment a TypeScript reader parses it.

28 new tests, plus three on the store's new reads. The route harness moved to a
`conftest.py` shared with the lease tests.

**Review round 1** — `9a29e73b`

- **The epoch fence was an accident of the heartbeat.** The lease was read
  before the log lock, so nothing stopped a batch committing under an epoch that
  had already moved — except that the trailing `renew` would then fail, and fail
  as an unhandled `LookupError` rather than an envelope. The lease is now read
  `FOR UPDATE` after the lock, so the generation a batch is checked against is
  still the one it lands under, and renew's failure is unreachable rather than
  handled. Ordered after the log lock deliberately: a lease claim takes only the
  row lock, so the two orders can never wait on each other.
- **A `session.upsert` may not name another host**, as well as another agent.
  §4 says the server checks agent, session *and* host against the lease; the
  host id was the one it did not check.
- **A position repeated inside one batch** was reported as a conflict with
  something that was not in the log — the batch's own earlier event. It is now
  compared against what that batch just wrote.
- The overlap read asked for the span containing the repeats rather than the
  repeats themselves, which for a host resending from a lost cursor is most of
  the log. One `IN` instead of two range predicates.
- The per-event 64 KiB cap is measured on the raw dict before parsing, replacing
  a sniff of the parse failure's message.

**One caveat on the proof.** The route-level fence test only discriminates
because it drives a batch that is *refused* — an accepted batch renews the
lease, and that UPDATE locks the same row, so a fence starting there would look
identical from outside. Named in the test's docstring. The store-level
`get_held` test is the one that pins the mechanism.

Round 2 found nothing to fix in the code. It did correct the lock-ordering
comment: a lease claim does not take only the row lock — it writes `sessions`
afterwards, as ingest does. The conclusion holds for a different reason, now
the one written down: every path that takes both takes `session_leases` first,
and the one path that inverts it registers a session that ingest cannot see
uncommitted. `2051745b`.

Six findings deferred, listed below.

</details>
---

<details>
<summary><b>S4 — Association and publication policy</b> ✅ landed</summary>

`audience` is gone from the contract, so a host no longer says who a thing is
for. `publication_policy(projection, association)` decides instead — a pure
function, and the only place the decision is taken.

```
publication_policy(projection, None)
  → None                                    nothing published, anywhere

publication_policy(projection, association)
  → PublicationPlan(
      room_id, thread_root_id,
      cards      = …requests still answerable (open, submitting)…,
      outcomes   = …requests already settled (resolved, closed)…,
      disclosures= …per-turn tool activity, transcript removed…,
    )
```

Associating is a person's decision, taken through the gateway:

```
POST /gateway/v1/sessions/{id}/room  {"roomId": "…"}
  → 200  {"sessionId", "roomId", "source": "granted", "grantedByActorId"}
  → 403  caller lacks manage on the agent, or write on the room
  → 404  session not registered, or room does not exist
  → 409  this session already publishes into a room
```

**Worth calling out**

- **`None`, not an empty plan.** A caller cannot iterate an empty plan's fields
  and quietly believe it did the right thing. The branch is forced.
- **The transcript exclusion happens before the renderer, not inside it.**
  `render_activity` draws a turn's messages *and* its tool disclosure in one
  call. What the model said and what a user typed into the Console are not a
  room's business — a room reply is something an agent chooses to say through
  `post_message`, and a session must not become a second mouth.
- **Two permissions, because the grant exposes two things to each other.**
  Manage on the agent behind the session, because publishing shows a room what
  that session is doing; write on the room, because it puts posts there.
  Neither implies the other, and a publicly writable room supplies only its own
  half.
- **The host cannot associate itself.** Nothing under `bridges/agent/` imports
  the service or the store, so "a leaseholder still cannot publish" is a fact
  about the code rather than about the checks currently written into it. A test
  scans the tree for it — there is no call to make, so there is nothing else to
  assert.
- **A turn started from a room does not quote that room back to itself.** An
  item whose origin names the target room with a message id is dropped from the
  disclosure; the association keeps it as the thread anchor.
- **409 comes out of a savepoint.** `associate` raises `SessionAlreadyAssociated`
  from inside `begin_nested`, so the caller's transaction survives the primary
  key violation and can still read which room won.

**Acceptance.** An unassociated session posts nothing, and nothing about the
session's own state can tip that. An associated one posts requests and tool
activity and no transcript text — checked against `examples.activity.json`
rather than hand-built items, with a guard test proving that recording really
does carry assistant speech, and against the Slack blocks the renderer would
actually draw. 28 new tests. Backed out both authorisation halves separately
and the transcript exclusion: each breaks tests that nothing else covers.

**Two calls I made that need your eye**

- The plan's *derive from a verified room command* path is **not built**.
  Nothing in the tree starts a session from a room message — `SessionDemo` is a
  stand-in and `_drop_session_command` discards answers — and no slice S1–S9
  introduces one. `source` records the distinction so a derived row has
  somewhere to go; the entry point waits for a caller.
- The gateway route is **not in the plan**. Without an entry point the granted
  path would be unreachable forever, so I added the smallest one. Say if you
  would rather it landed with S9's read routes.

**Review round 1** — `e3709e52`

- **The echo rule could not fire for the case it described, and broke the one
  it could reach.** `_publishable` suppressed any item whose origin named the
  target room. Past the transcript guard the only kind left is `tool-activity`,
  so the room's own message — a `user-message` — never got there; what the rule
  actually did was empty a room-steered turn's disclosure, and an empty
  disclosure is dropped, so the room that asked for the turn saw nothing of it.
  Being caused by a room is not being said by it. Rule deleted, the model
  comment claiming it corrected to what `origin_message_id` really does (the
  thread anchor), and two tests where there were two.
- The tree scan now knows the model's name as well as the store module and the
  service — `SessionRoomAssociation` contains neither of the other two spellings
  — and its docstring says it is a tripwire on the likeliest accident, not a
  proof.

Five findings deferred, listed below.

</details>

---

## Deferred findings

Raised in review, judged not worth blocking the slice, to be swept in a pass at
the end of the branch.

| From | Finding |
|---|---|
| S3 | Repeat matching compares bodies. When `audience` is dropped from `contract.py` a host resending a pre-upgrade event will be told `IDEMPOTENCY_CONFLICT` for content that has not changed. Event-id equality is checked first, so the body comparison only adds detection of a host reusing an id — decide whether that is worth the upgrade hazard. |
| S3 | `uq_session_events_sequence` is the one constraint `append` does not name. If the other two are worth named exceptions as a backstop, so is that one; if not, neither is. |
| S3 | The 1 MiB batch cap is enforced after the whole body is buffered and parsed, so the memory it is there to bound has already been spent. |
| S3 | `record_reported_state` can raise an uncaught `LookupError` — the other half of the renew finding. Unreachable while nothing deletes a session row. |
| S3 | A lease claim now blocks on an in-flight batch, by design, and app connections set no `lock_timeout` (`engine.py` sets only `idle_in_transaction_session_timeout`, which does not fire on a lock wait). Bounded by the 1 MiB batch cap today; revisit if a later slice puts anything slow inside that transaction. |
| S4 | `_disclosures` iterates `snapshot.turns`, so an item whose `turn_id` has no matching turn is never published. `SessionProjection.apply` folds an `ItemUpsert` without requiring the turn to exist, so that state is reachable, and the drop is silent. |
| S4 | `expires_at` is read nowhere. `LIVE_STATES` makes a request whose deadline has passed a card with live buttons. |
| S4 | "Answerable" is written three times — `LIVE_STATES`, `projection.open_requests()`, and the `pending_request_ids` literal in `projection.py`. Adding a state updates two of three. |
| S4 | Nothing drives the association route through the app: the tests call the handler as a function, so the path string, `include_router`, the dependency wiring and `get_current_user`'s 401 are unexercised. Each checked by hand at this ref. Regression cover, wanted before the branch merges. |
| S4 | The route answers 200 for a create where the gateway's other creates answer 201. Worth changing with the app-level test above, not before it — called as a function, the status code is invisible. |
| S2 | `INVALID_REQUEST` and `LEASE_HELD` are two codes beyond the contract's §6.2 table. Flagged above; needs a contract decision, not a code change. |
| S2 | Operators on the default `ingress.mode: existing` need a manual manifest edit for `/agent` to reach the agent API. Wants a release-notes line. |


amaudruz:
Closing unmerged: CHOO-2667 is closed as a duplicate — the work is largely covered by #419 (worktree-sdk-providers). Branch left in place for reference.

PR #412, opened 2026-09-09
Removing an agent from a room does not stop message delivery

jameshoweee:
Removing an agent from a room deletes the membership row, but a running client holds its own per-room subscription and keeps receiving everything said in the room. Matrix used to enforce the kick at the homeserver, and that enforcement went away with it.

Same leak on the pull side: `poll_events` returns everything buffered for the agent with no membership check, unlike its sibling `poll_room_events`.

Fix: a removal is rung over the invite bus the same way an invitation is, and the transport drops the room's subscription and its cursor. `poll_events` filters by the rooms the agent is in.

Tests fail before the change and pass after.

### DIFF
--- a/core/switch_core/bridges/agent/protocol/event_buffer.py
+++ b/core/switch_core/bridges/agent/protocol/event_buffer.py
@@ -255,8 +255,19 @@ class EventBuffer:
     # Confirming clients get at-least-once instead.
     # ------------------------------------------------------------------
 
-    async def poll(self, agent_id: str, timeout: float = 30) -> list[AgentEvent]:
-        return await self._legacy_poll(agent_id, "legacy:all", timeout=timeout)
+    async def poll(
+        self, agent_id: str, timeout: float = 30, rooms: set[str] | None = None
+    ) -> list[AgentEvent]:
+        """Everything queued for this agent, optionally limited to `rooms`.
+
+        `rooms` is how the caller applies membership. The buffer is keyed by
+        agent and knows nothing about who is in what, so an event queued while
+        the agent was a member stays queued after it is removed; passing the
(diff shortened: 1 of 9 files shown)


christian-mcdermott (approved):

core/switch_core/bridges/agent/protocol/event_buffer.py
```diff
@@ -280,20 +339,39 @@ async def _legacy_poll(
...
+
+        Past everything *scanned*, not merely everything returned. A legacy
+        read has no limit, so it always reaches the end of the buffer, and
+        events the filter excluded are events this reader will never want. Left
+        behind them the cursor would park below the head indefinitely; when
+        retention eventually trimmed them the next read would raise
+        `CursorExpiredError` and log that the poller missed events, which would
+        be a false alarm — they were filtered, not lost.
+        """
+        scanned_through = self.head(agent_id)
```
petr-sandbox: Should the room-filtered events be excluded from `scanned_through`?

Scenario: `server_connectors/core.py:223` polls with `timeout=30`. At T0 membership is `{A}`; at T5 the agent is added to room B and addressed there. The event wakes the poll, gets filtered out by the stale `rooms` set, and the cursor is confirmed past it. Next poll sees `{A, B}` but starts after that seq — the addressed message and the `room_join` never arrive. Reproduced against PR head:

q.enqueue("a1", "B", msg)
poll(rooms={"A"})      -> []
poll(rooms={"A","B"})  -> []      # retained forever
q._cursors["a1"] == {'legacy:all': 1}

petr-sandbox (approved):
Looks good. Just one thing my Claude found.

PR #421, opened 2026-09-09
feat(templates): template registry — server-side storage and API (CHOO-2677)

christian-mcdermott:
## What

A Switch server can now hold templates. Until now a template was a YAML file passed around by hand; this is the storage and the API under it, so one can be uploaded once and found by everyone on the server afterwards.

Upload, list, search by name or description, fetch back, edit, and delete your own — plus a Templates tab on the dashboard to do all of it.

## Shape

**Storage.** A `templates` table: `id`, `owner_id`, `name`, `description`, `kind`, `content`, `version`, `created_at`, `updated_at`.

Two choices are load-bearing, and both exist to keep this table still while the format keeps moving:

- **`content` is stored verbatim and never parsed.** That is what makes the round trip byte-identical, and it means a document written for a newer format than this server understands is stored and served back intact rather than refused at the door.
- **`kind` is free text**, not an enum or a check constraint. Room, group and agent templates should differ by a string, not by a migration.

Uniqueness is `(owner_id, name)`, so two people may each keep a "deploy-room" but neither may keep two. Scoping it to the owner rather than the server is also what lets it survive the tenant work: an owner belongs to a tenant, so the constraint narrows instead of needing to be rebuilt.

**Access.** The catalogue is server-wide — a registry nobody else can see distributes nothing — so every owner's templates are listed to everyone. Ownership decides who may *change* a template, via the existing `require_manage`, so an admin can clean up and nobody else can touch what is not theirs.

**API.** `GET /templates` (with `q`, `kind`, `owner_id`), `POST /templates`, `GET /templates/{id}`, `GET /templates/{id}/content`, `PATCH /templates/{id}`, `DELETE /templates/{id}`.

Two doors for reading: the detail route returns the document inside a JSON envelope with its metadata, and `/content` returns it alone, byte for byte, for anything that wants to pipe it somewhere. A listing omits the document bodies.

Uploads are bounded by `TEMPLATE_MAX_BYTES` (1 MiB default), counted in encoded bytes rather than characters so multibyte text cannot slip past, and refused whole rather than truncated. The column is unbounded `Text`, so raising the limit is a deploy-time change and never a migration.

**UI.** A Templates tab on the Resources page — catalogue with search and filters, an upload dialog that can load from a file, and a detail page to read, edit, copy, download or delete. Copy and download re-fetch from the server rather than using what is on screen, so what lands in the file is what is stored even with unsaved edits in the editor.

## Acceptance

- Uploaded, listed, searched and fetched back **byte-identical** — proven over real HTTP, with a document carrying CRLFs, tabs, trailing spaces, quotes, backslashes and multibyte characters.
- Templates from different owners are all visible in the listing.
- Deleting someone else's is refused (403) and the template survives.
- New tests green; existing suite green and unchanged — **2746 passed**.

## Notes for review

- **Nothing in `rooms_yaml.py` is touched.** `CHOO-2665`, `CHOO-2656` and `CHOO-2658` are all editing that file and `gateway/rooms.py` right now; this branch stays out of both. Storage is opaque to the format by design.
- **No agent-facing MCP tools**, so the three connector skills are untouched and `test_mcp_tool_surface.py` is unaffected.
- **No tenant column.** Single-tenant today; when the tenant sweep lands it has to touch every owned table in one revision, and `templates` is one more line in it.
- **A real bug surfaced in review of my own tests**: `updated_at` is computed by Postgres, so a flush leaves the attribute expired and reading it back raised `MissingGreenlet` instead of returning a timestamp — on every edit, in production, since `expire_on_commit=False` is the same there as in tests. The store now reloads the row, with a regression test at the layer where it lives.
- The **Console** Templates tab (on `work/capture-export`) is file-based today and has no registry behind it. Pointing it at this API is a separate piece of work — deliberately not done here to avoid colliding with that branch.

## Not in scope

The Console marketplace section and its UX, pre-bundled out-of-the-box templates, and any public or cross-server marketplace.


core/switch_core/gateway/templates.py
```diff
@@ -0,0 +1,274 @@
...
+) -> Response:
+    """The stored document itself, byte for byte, with no envelope around it."""
+    template = await template_store.get(session, template_id)
+    if template is None:
+        raise HTTPException(
+            status_code=404, detail=f"Template not found: {template_id}"
+        )
+    return Response(
+        content=template.content.encode("utf-8"),
+        media_type="application/x-yaml",
```
abeldantas: Can we keep this as `text/plain` to not force to yaml?
I think future templates might not all be yaml.
thanks

amaudruz (approved):

PR #422, opened 2026-09-10
Multi-tenancy Phase 1: the tenant model (CHOO-2623)

wojtyniakAQ:
Phase 1 of multi-tenancy, schema only. The existing deployment becomes tenant
zero and nothing changes for anyone using it.

The design, with the reasoning behind each choice and an honest list of what
this phase deliberately does not close, is committed alongside the code:
`docs/old/multi-tenancy-phase1-db.md`. It is worth reading before the diff —
the diff is large and mechanical, the design is where the decisions are.

## What is here

- `tenants` and `tenant_members`. Membership is a row, not a column on the
  user, because a person may belong to more than one tenant.
- `tenant_id` on the 36 tables holding customer data. Four tables stay global:
  `users`, `oidc_identities`, `feature_flags`, `alembic_version`.
- Per-tenant uniqueness where a global constraint would break on the second
  customer — agent names most obviously.
- Composite foreign keys between scoped tables.
- One migration turning the existing deployment into tenant zero.

## What is not here

- **Row-level security policies** — the next change.
- **Session plumbing**, resolving the tenant from the caller — the change after
  that.
- **The runtime database role** — split out as its own ticket. Until it lands
  the service still connects as a superuser, and a superuser ignores row-level
  security outright, so the policies that arrive next will not bite in a
  deployed environment. Safe while one tenant exists; a hard prerequisite for
  the second.

## Two things reviewers should look at hardest

**Composite foreign keys are the expensive half of this.** Referential
integrity checks in Postgres deliberately bypass row-level security, so a
policy cannot stop a row referencing another tenant's parent — only the key
can. The cost is 53 keys rewritten and 13 extra unique indexes. If that trade
looks wrong, that is the thing to argue with.

**The five keys that null on delete name their column explicitly**
(`ON DELETE SET NULL (group_id)` and friends, Postgres 15+). A plain
multi-column SET NULL nulls `tenant_id` as well, and the delete then fails on
the not-null constraint. This was caught in review rather than in production;
one of the five is covered by an explicit check described below.

## Locking: measured against production-sized data

**Verdict: a blink, not an incident. No change needed.**

Measured by restoring a copy of the pilot database into a throwaway container
and running the migration against it with lock logging on:

- **Total migration time: 0.92s.** Downgrade 0.69s, data intact — row counts
  identical before and after.
- **Longest any table is unwritable: ~0.6s.** The whole revision runs in one
  transaction, so every `ACCESS EXCLUSIVE` lock is held to commit; the
  transaction span is ~620ms and that is the real answer, not the per-statement
  times.
- Slowest single statement is the unique index on `messages` at 44ms. That is
  the one that scales with data, so it is the one to watch.
- Disk cost of the new columns and indexes: ~8.4MB on a 291MB database.

Adding the tenant column is metadata-only and does not rewrite the table, which
is what keeps this cheap. The thirteen unique index builds are the part that
grows with data.

**When to revisit — now measured, not extrapolated.** The 100x figure above
was a guess from one data point. Since then we generated synthetic databases
at 1x, 5x and 20x the pilot's row counts
(`core/tests/synthetic/generate_synthetic_db.py --scale-factor`) and reran
the same lock-logged migration against each in a throwaway container:

| scale | messages | transaction span | slowest statement (unique index on `messages`) | total wall time | disk growth |
|-------|----------|-------------------|--------------------------------------------------|------------------|-------------|
| 1x    | 63k      | 1.0s               | 53ms                                             | 1.6s             | +5.6MB      |
| 5x    | 314k     | 1.6s               | 270ms                                            | 2.7s             | +25MB       |
| 20x   | 1.26M    | 3.4s               | 892ms                                            | 3.8s             | +97MB       |

The slowest statement — the `messages` unique index build — scales almost
exactly linearly with row count, as expected for a btree build. But the
*transaction span* grows much slower than the row count: 20x the messages
only triples the lock time, because roughly 900ms of every run is fixed
cost — the other ~35 scoped tables each get a handful of small,
constant-cost `ALTER TABLE` / `CREATE POLICY` statements, independent of how
many rows they hold.

That is a straight line, not a knee: there is no point where this suddenly
stops being a blink. Extrapolating the measured per-row cost to 100x the
pilot's size (~6.3M messages) predicts roughly 13 seconds of exclusive lock
on `messages` — a real cost, and a predictable multiple of what we measured,
but not the cliff the original claim implied. The more useful takeaway is
that a 20x deployment, not an extreme one, already spends 3-4 seconds unable
to write, which is worth planning around well before 100x. **Revised
guidance:** keep this as one transaction for now; revisit `CREATE INDEX
CONCURRENTLY` once total message volume heads into the low millions (roughly
30-50x today's pilot size), where the lock span starts running several
seconds rather than one.

The `NOT VALID` / `VALIDATE CONSTRAINT` split on the foreign keys buys nothing
inside a single transaction. It is kept because it is the shape a future
two-phase split would need, and commented as such.

## How it was verified

Not against an empty database, which proves little. The migration was run over
a pre-change schema holding real rows, then reversed, then re-applied:

- Row counts preserved; every scoped table carries tenant zero afterwards.
- `tenant_members` backfilled from existing accounts, admins as owners.
- All rewritten keys validated; no constraint left `NOT VALID`.
- Deleting a client nulls only the sender column and leaves the tenant intact
  — the exact failure the column-list form exists to prevent.
- `downgrade` returns constraints and indexes byte-for-byte to the baseline.
- Full suite green: 2691 passed, 6 skipped.

One pre-existing failure in the integration suite
(`test_agent_greeting_e2e.py::test_no_greeting_when_the_bridge_has_them_disabled`)
was reproduced on `main` with none of these changes present, so it is unrelated
and left alone.

### DIFF
--- a/core/switch_core/bridges/agent/protocol/service.py
+++ b/core/switch_core/bridges/agent/protocol/service.py
@@ -337,8 +337,9 @@ class ProtocolService:
             InvalidIconUrl: ``icon_url`` is malformed or points somewhere unsafe.
             InvalidDisplayName: ``display_name`` is over-long or unsafe to render.
             AgentExistsError: agent with this name exists and ``overwrite`` is
-                False, or the existing agent is owned by another user (the name
-                is global, but re-registration stays within the owner's tenant).
+                False, or the existing agent is owned by another user
+                (``agents.name`` is unique per tenant, but re-registration
+                stays within the owner's tenant).
         """
         if not _VALID_NAME_RE.match(name):
             raise ValueError(
@@ -367,12 +368,13 @@ class ProtocolService:
                     "replaces integration profile)."
                 )
             if existing and existing.owner_id != owner_id:
-                # Agent names are a global namespace, but re-registration must
-                # stay inside the caller's own tenant. Overwriting an agent
-                # owned by someone else would mint the caller a live API key for
-                # that agent (whose owner_id is left unchanged) and delete the
-                # real owner's key, i.e. a cross-tenant takeover. Report it as a
-                # plain name clash so ownership is never disclosed.
+                # `agents.name` is unique per tenant, not globally, but
+                # re-registration must still stay inside the caller's own
+                # tenant. Overwriting an agent owned by someone else would mint
+                # the caller a live API key for that agent (whose owner_id is
+                # left unchanged) and delete the real owner's key, i.e. a
+                # cross-tenant takeover. Report it as a plain name clash so
+                # ownership is never disclosed.
                 raise AgentExistsError(
                     f"Agent already exists: {name!r}. "
                     "Pass overwrite=True to re-register (rotates API key, "
--- a/core/switch_core/db/models.py
+++ b/core/switch_core/db/models.py
@@ -8,6 +8,7 @@ from sqlalchemy import (
     Column,
     DateTime,
     ForeignKey,
+    ForeignKeyConstraint,
     Index,
     Integer,
     LargeBinary,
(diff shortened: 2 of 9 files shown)


petr-sandbox:
Quick review from my Claude (not mandatory):
---
- `tenant_members` is written by the migration and never again. Four user-creation paths write `users` alone. On a fresh install the migration runs before any user exists, so the table stays empty forever while an upgraded deployment has rows — same version, different data. The next PR resolves a gateway user's tenant from exactly that table, so everyone created in between is locked out. Cheap now, nasty later.
- `reference_type_store.create()` 500s on the second tenant. A `scalar_one_or_none()` inside the `IntegrityError` handler raises from inside the handler, turning a clean "already exists" into a 500. One word. The file's already open in this PR. Same store's `get/update/delete` hardcode tenant zero while `list_*` are unscoped — so it hides rows that exist and shows rows that aren't yours, disclosed only by a TODO.
- The new catalogue test fails open. It derives "scoped" as has a `tenant_id` FK, so a new table with customer data and no `tenant_id` simply isn't in the set and nothing fails — the exact regression the design says these tests exist to prevent. Invert it: hardcode the five global tables, assert everything else is scoped.

wojtyniakAQ:
Thanks — the third one is a real bug, and it was worse than you described.

**The catalogue test failing open: correct, and mine.** I specified it the wrong
way round: I told it to derive the scoped set from the models "so a table added
later cannot be silently missed", which is exactly backwards, since a table with
no tenant column is then not in the set at all.

Proved rather than reasoned about: I added an `invoices` table with customer data
and a `room_id`, no tenant column, and ran both catalogue files. 14 passed.
Nothing noticed.

Inverted as you suggested — the global list is hardcoded and every other table in
the metadata must be scoped, have a policy, and carry the tenant on its keys. Both
catalogue tests now share one derivation instead of keeping private copies, which
was the other way they could quietly disagree. With the fix, the same experiment
kills the suite at import, naming the table.

One correction to your count: it is three global tables, not five. `alembic_version`
is Alembic's and never registers on the metadata, so it needs no entry.

**The other two were true of this PR and stale at the top of the stack** — you were
reading the bottom of four. Verified by running it, not by reading diffs:

*`tenant_members` diverging between a fresh and an upgraded install.* Built both
against a real database — empty → migrate → seed, and pre-tenant → populate →
migrate → seed — and compared. Both end with every user holding a membership and
no orphans. All four user-creation paths funnel through one place that writes the
membership row, added in the very next commit after this PR. Already pinned:
deleting that write turns six tests red.

*`reference_type_store.create()` 500ing on a second tenant.* The clash lookup does
name the tenant as of the top commit, so the raise-from-inside-the-handler is gone —
but you are right that nothing pinned it, and nothing *could* have, because
reproducing it needs the same slug alive in two tenants at once and no test had two.
Added one. Removing the tenant predicate reproduces your 500 exactly
(`MultipleResultsFound` from inside the `IntegrityError` handler). The
`get`/`update`/`delete` hardcoding and the unscoped `list_*` are also resolved at the
top — all nine methods resolve the caller's tenant now.

Suite is 2833 green. The catalogue fix is the only one of the three that logically
belongs on this PR; it is committed at the top of the stack rather than here, to
avoid rebasing four branches for one test file.

petr-sandbox (approved):
Looks good.

PR #423, opened 2026-09-10
Multi-tenancy Phase 1: bind the tenant to every session (CHOO-2623)

wojtyniakAQ:
Stacked on #422 — review that one first; this diff is against it, not against
main.

Second of four. #422 gave every table a tenant column. This one works out
which tenant a caller belongs to and puts it on the database session. Still no
row-level security, so still no behaviour change: nothing yet reads the value
this sets.

## The one idea

The tenant is set by a SQLAlchemy `after_begin` event hook, not by a call at
each session site. That is deliberate and it is the whole point of the change:
it makes it structurally impossible to open a session that forgets. Two of the
four prior-art bugs this design was built against are a call site that forgot,
and the value then surviving on a pooled connection into the next request. The
hook closes the first; `is_local=true` closes the second, because the setting
is released at commit.

There is a test that reuses the same pooled connection — asserted by backend
pid — and proves the tenant does not carry over.

## Where the tenant comes from

Never from a request parameter.

- **Gateway:** the caller's membership row. Exactly one tenant exists today, so
  resolution returns the single membership and **raises if there are none or
  more than one** rather than picking. Guessing here is how you get a
  cross-tenant read.
- **Agent bridge:** the api key's tenant, bound in middleware, which runs
  before dependencies.

Both lookups necessarily happen before a tenant is known, so they run on an
explicitly-named unscoped session rather than an incidentally unscoped one.

## Two things worth a reviewer's attention

**No endpoint call site changed.** ~107 of them take a session dependency that
resolved before authentication, which looked like it needed rewriting. It does
not: constructing the session does no I/O, the transaction begins on the first
query inside the endpoint body, and every dependency has resolved by then.
Verified that no other dependency queries the database during its own
resolution. If that reasoning is wrong, this is where it is wrong.

**Creating a user now creates its membership row too.** Not in the original
scope, but a necessary consequence of resolution refusing to guess — otherwise
every account created after this lands could never sign in. Applies equally to
admin-created, JIT-provisioned and seeded accounts.

## Not here

- Row-level security — next change.
- The ~206 background session sites — the change after that. Until then those
  run with no tenant bound, which is harmless while nothing enforces.
- The insert default still falls back to tenant zero when nothing is bound, for
  the same reason. The fallback goes when the background sites are converted.

## Verified

2720 passing, up from 2704, 16 new tests. `check` and `typecheck` clean. The
new tests cover the hook setting and releasing per transaction, pooled reuse
not leaking, a real gateway request over ASGI binding the caller's tenant, a
bearer request binding the api key's tenant, and resolution raising on nought
or two memberships.

One pre-existing integration failure was reproduced on the unmodified baseline
and left alone.

### DIFF
--- a/core/switch_core/bridges/agent/auth.py
+++ b/core/switch_core/bridges/agent/auth.py
@@ -16,6 +16,7 @@ from switch_core.db.models import Agent, ApiKey
 from switch_core.db.stores.agent_store import AgentStore
 from switch_core.db.stores.api_key_store import ApiKeyStore
 from switch_core.logging_context import bind_log_context, unbind_log_context
+from switch_core.tenant_context import bind_tenant_id, unbind_tenant_id
 
 logger = logging.getLogger(__name__)
 
@@ -82,6 +83,17 @@ class BearerAuthMiddleware:
     Public paths (see ``PUBLIC_PATH_PREFIXES``) bypass authentication
     entirely. The MCP path also requires an agent — registration tokens
     are not enough to open an MCP session.
+
+    This is where an authenticated request's tenant gets bound (see
+    ``switch_core.tenant_context``), not a FastAPI dependency: this class runs
+    as ASGI middleware, ahead of routing and dependency injection entirely,
+    so there is no ordering question about whether a downstream ``get_session``
+    might query before the tenant is known — it cannot, since it does not run
+    until after ``self.app(...)`` is called below. The lookups that resolve
+    the credential itself (``_resolve_api_key``, ``_try_oidc``) run on their
+    own session, opened directly from ``session_factory`` rather than through
+    that downstream seam, because they are what determines the tenant and so
+    must run before one is bound.
     """
 
     def __init__(
@@ -134,11 +146,18 @@ class BearerAuthMiddleware:
         if agent is not None:
             scope["agent"] = agent
             scope["agent_id"] = agent.id
-            token = bind_log_context(agent_id=agent.id)
+            # api_key.tenant_id is the source of truth (docs/old/
+            # multi-tenancy-phase1-db.md, "Setting the tenant"); the OIDC
+            # path resolves straight to an Agent with no ApiKey row, so it
+            # falls back to the agent's own tenant.
+            tenant_id = api_key.tenant_id if api_key is not None else agent.tenant_id
+            log_token = bind_log_context(agent_id=agent.id, tenant_id=tenant_id)
+            tenant_token = bind_tenant_id(tenant_id)
             try:
                 await self.app(scope, receive, send)
             finally:
-                unbind_log_context(token)
+                unbind_tenant_id(tenant_token)
+                unbind_log_context(log_token)
             return
 
         # Registration token: pass through (handler validates again). MCP rejects.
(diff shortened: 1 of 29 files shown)


petr-sandbox:
Some nitpicks:
---
- `_tenant_id_default` falls back to tenant zero with no log at all. Proven: a background insert lands in tenant zero and `caplog` captures nothing at WARNING or above. That's the repo's own priority-4 "silently degrades to look fine". A `logger.warning` costs nothing and is the difference between a discoverable and undiscoverable migration later.
- The two fallbacks disagree. A background write logs `tenant_id=default` and writes `tenant_id=00000000-…`. Point the logging filter at `TENANT_ZERO_ID`.
- The new comment in `server_connectors/core.py` says `_register_agent` runs "with no tenant bound". False on `POST /connectors` — a tenant is bound there. Doesn't matter today, but the next reader will lean on it.
- `_seed_admin_user` early-returns when the admin exists, so it never repairs an admin created on #422 code after that migration ran. No endpoint writes a membership either (`TenantMemberStore.create` is dead code), so a password-login user is unrepairable in-app — a long-lived staging box on the branch gets a 403 wall and needs SQL. Either repair on the existing-admin branch or ship both PRs as one release.

wojtyniakAQ:
Four findings, three real, one already dead. Verified at the top of the stack by
running rather than reading — you were looking at this PR alone and it is four
deep now.

**Silent tenant-zero fallback — dead.** `_tenant_id_default` no longer exists;
the top commit renamed it to `require_tenant_id` and made it raise. Your point was
right about this PR's tree, and it is what motivated removing the fallback, which
in turn exposed eight junction tables that had been ignoring the caller's tenant
entirely. So it paid off twice.

**The two fallbacks disagreeing — real, and not for the reason either of us
thought.** The log line was not stale-fallback drift: `LogContextFilter` never
consulted the tenant context at all, only the per-request log context. So *any*
background tenant binding was invisible to logging, not just tenant zero's. The
filter now reads request context, then the bound tenant, then a placeholder.

I deliberately did not point the config default at tenant zero's uuid, which is
the obvious fix. That would make every unattributed line indistinguishable from a
real customer's when someone filters by tenant. It stays a placeholder, and a test
pins that it is not a tenant id.

**The false comment — real, wrong file.** `server_connectors/core.py`'s comment is
accurate; the wrong one is `register_agent_with_token`'s docstring in
`protocol/service.py`, which *names* core.py, which is presumably how it found
you. It was wrong twice: a tenant is bound on `POST /connectors` as you say, and it
also still claimed rows would fall back to tenant zero, which stopped being true
two commits later. Rewritten.

**The unrepairable account — real, and the window is precise.** The migration
backfills everyone alive at migration time; `UserStore.create` only began writing
memberships in the second commit of this PR. Anything created on a box running the
previous PR's code, after that migration, has none. Only the OIDC identity-link
path repaired it, so a password-login account had no in-app remedy — exactly your
403 wall.

Fixed by repairing on the existing-admin path at startup, logging only when it
actually writes. You were also right that `TenantMemberStore.create` is dead code:
zero callers anywhere including tests. Removed, and the class now says why there is
deliberately no second way to write a membership.

Your alternative — ship the stack as one release — is still the safer call and I
have put it to Wojciech. The repair path means we do not have to rely on that
holding.

2862 green. Each fix pinned by reintroducing its defect.

petr-sandbox (approved):
Looks good.

PR #424, opened 2026-09-10
Multi-tenancy Phase 1: bind a tenant to background work (CHOO-2623)

wojtyniakAQ:
Third of four, stacked on #423, which is stacked on #422. Review those first;
this diff is against #423.

#423 bound the tenant on the request paths. This does the session sites with no
request behind them — the delivery loop, the bridges, room service, startup.
Still no row-level security, so still no behaviour change; this is the last
change before anything enforces.

## The rule: nothing is ambient

An earlier version of this change let a long-lived task pick up its tenant once
at creation and keep it for its lifetime. That was wrong in four separate ways,
so the rule is now:

> Every unit of background work binds the tenant of the row it is acting on, at
> the point it acts. Long-lived tasks bind nothing — each one clears the tenant
> as its first act. A lookup that must span tenants uses a helper that
> genuinely unbinds.

The four leaks that rule removes, rather than patches:

- Restarting a bridge from the admin UI spawned its task under the **requester's**
  tenant, and it stayed there for the bridge's whole life.
- A puppet client is reused across every room on a bridge; its task kept the
  first room's tenant forever.
- A room created after its bridge started missed the cache and fell back to
  whatever was ambient.
- **Mattermost has no ambient context at all** — its inbound arrives on an OS
  thread via `run_coroutine_threadsafe`, which snapshots an empty one. Auto-created
  rooms would have landed in tenant zero there and in the bridge's tenant on
  Slack. A per-platform lottery is a sign the model is wrong, not the platforms.

Platform is now irrelevant, pinned by a test that dispatches from a real OS
thread.

## The escape hatch now does what it says

`unscoped_session` previously inherited whatever was bound and the hook stamped
it — the name was false and the allowlist test certified it. It clears the
tenant for the block and restores it after. The audit is now two tests: which
modules may call it, and which may open a raw session factory at all.

## Two things worth a reviewer's attention

**A concern that turned out to be unreachable, and why that is good news.**
Review flagged that the system client is a member of every tenant's rooms, so
scoping its enumeration would make it silently deaf. It is not: `client_rooms`
carries composite keys to both parents on the tenant, so a client *cannot* be a
member of another tenant's room — verified with a probe insert that fails with
a foreign key violation. That is #422's composite keys doing exactly what they
were argued for. The ambient-inheritance bug underneath was real and is fixed;
the schema fact is now pinned by a test rather than argued in a PR.

**A test that could not be written the obvious way.** Without policies, a
scoped read still returns every row, so "does this leak across tenants" cannot
be asserted as row visibility yet — the first attempt passed with the defect
reintroduced. Those tests assert on the bound tenant and on `app.tenant_id`
instead. They convert to row-visibility assertions once the next change lands.

## Cost, measured

The hook adds one `set_config` round trip per transaction: **0.20–0.36 ms**
against a 0.95–1.20 ms baseline, over 400 single-statement transactions per arm.
That is a test container over a local socket, so treat it as an upper bound on
shape rather than a production number. On the delivery path it is roughly per
message once a handler writes.

## Found, out of scope, tracked as CHOO-2687

Registering an agent in one tenant creates a bot carrying its name and
description in **every** tenant's Slack, Mattermost and Teams workspace —
`_create_bridge_identities` iterates all bridges regardless of tenant. It
predates this work, and no database session is involved, so neither this change
nor row-level security will catch it. Filed separately; not fixed here.

## Not here

Row-level security, and removing the tenant-zero fallback from the insert
default. Both next.

## Verified

2799 passing, up from 2744 across the two commits. `check` and `typecheck`
clean. Each of the eleven regression tests was checked by reintroducing its
defect and confirming the failure.

### DIFF
--- a/core/switch_core/bridges/agent/protocol/service.py
+++ b/core/switch_core/bridges/agent/protocol/service.py
@@ -72,7 +72,7 @@ from switch_core.db.models import (
     Tool,
     User,
 )
-from switch_core.db.session_scope import tenant_session, unscoped_session
+from switch_core.db.session_scope import unscoped_session
 from switch_core.db.stores.agent_runtime_state_store import (
     IDLE as RUNTIME_STATE_IDLE,
 )
@@ -475,7 +475,12 @@ class ProtocolService:
         the source of truth, that guess would confirm itself forever.
         """
         token_hash = hashlib.sha256(registration_token.encode()).hexdigest()
-        async with self.session_factory() as session:
+        # Resolving a credential is unscoped by definition — `key_hash` is one
+        # of the two columns deliberately left globally unique for exactly
+        # this reason — and it is the read that *produces* the tenant bound
+        # below. Answering it under a tenant would make it either a tautology
+        # or a false "invalid token".
+        async with unscoped_session(self.session_factory) as session:
             key = await self.api_key_store.get_by_hash(session, token_hash)
             if key is None or key.type not in REGISTRATION_KEY_TYPES:
                 raise PermissionError("Invalid registration token")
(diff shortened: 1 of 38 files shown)


petr-sandbox:
More nitpicks:

`room_service.py` is the largest file in the PR (+295) and its tenant rework is completely untested. Proven: shadow `tenant_scope/tenant_session/unscoped_session` with no-ops inside that one file and the entire suite still passes, 2799. The three touched room-service tests only add `tenant_id` fields to fakes so they stop raising `AttributeError` — zero new assertions. Everywhere else in this PR got real contextvar assertions; this didn't. A later refactor dropping the bind from `delete_room` or `change_bridge` is green in CI.

- `reconcile_room_clients` aborts boot the day a second tenant owns a room. Reproduced against live Postgres: one tenant-zero admin client, one tenant-B room → `ForeignKeyViolationError` on f`k_client_rooms_client`, no per-room try/except, propagates out of `run()` and kills startup for everyone including tenant zero. Pre-existing rather than introduced — but this PR converted that exact call site, and your own new test's docstring says "the system client is one row per tenant, which is what the Phase 1 design says it becomes". Nothing at startup makes that true; `ensure_system_client` creates one row in tenant zero and early-returns forever.
- `server_connectors/lifecycle.py:164 remove()` is the one lifecycle verb left unconverted while `start/start_all` were done. Post-RLS a tenant-A operator deleting a tenant-B connector matches zero rows and still logs "Removed server-side connector" — silent false success, category 4 under the repo's own philosophy.

wojtyniakAQ:
All three real. Two were worse than you wrote, and the room-service one is the
best catch anyone has made on this stack — including four adversarial review
passes of my own that all missed it.

**Room service untested.** Reproduced your experiment at the head of the stack:
helpers shadowed, **2833 passed** — byte-identical to the baseline, not one test
moved. You are right that the three touched tests only stop fakes raising
`AttributeError`.

New tests assert `current_setting('app.tenant_id')` on the session the service
hands the store, rather than asserting on rows — row assertions are blind here,
because we still connect as the tables' owner, so an unbound session writes just
as happily as a bound one. **Your experiment is now the acceptance criterion**:
with the three helpers no-op'd, 5 fail. Thank you for it; I have adopted it as
the way to check the other reworks in this stack are load-bearing.

**`reconcile_room_clients` aborting boot.** Reproduced at the head:
`ForeignKeyViolationError` on `fk_client_rooms_client`,
`Key (tenant_id, client_id)=(tenant-B, admin-zero)`. Worse than you found in two
ways. The top commit's change to create one admin client per tenant did not close
it — it made it worse, because the registry now holds several and every room was
being offered all of them. And **`create_room` has the identical defect**, which
your report did not reach. `get_by_type` now requires a tenant; reconciliation
fans out per room with failures contained and reported rather than killing boot.

That containment is deliberately broader than this repo's "catch-and-log only in
event loops" rule, and I want it on the record: this runs inline before the
server binds, so an escape is not one skipped room, it is the whole deployment
refusing to start over one tenant's data. It logs per room and reports a count.
Say if you would rather it propagate — the tenant filter is the actual fix and
stands without it.

**`remove()` unconverted — and the live half is the other way round.** Under a
restricted role you get the silent false success you described. Under the owner
role **the deployment actually runs today**, the cross-tenant delete *succeeds*.
That is a live bug, not a post-RLS one. The store now raises rather than reporting
zero rows, and the visibility check moved ahead of teardown, because the core
registry is process-wide and the old ordering let a wrong-tenant caller stop
another tenant's connector and delete its agents before the delete failed.
Checked the sibling verbs: `start`/`start_all` correct, `stop`/`stop_all` touch no
database, `register` inherits correctly.

2862 green. Each fix pinned by reintroducing its defect.

petr-sandbox (approved):
Looks good.

PR #425, opened 2026-09-10
Multi-tenancy Phase 1: turn on row-level security (CHOO-2623)

wojtyniakAQ:
Last of four, stacked on #424 → #423 → #422. Review those first.

This one grew well past its original scope. It was "turn the policies on"; it
is now "turn the policies on **and make them apply**", because turning them on
without the second half produces a schema full of decoration.

## What it does

**Row-level security**, one policy per scoped table, `USING` and `WITH CHECK`
both, keyed on a `require_tenant_id()` that raises when nothing is bound.

**The service connects as a restricted role** — not the owner. Without this the
policies never bite, because Postgres exempts a table's owner. This was
originally split out as its own ticket; it came back in because the phase means
nothing without it.

**Seven `SECURITY DEFINER` functions as the entire exemption.** Resolving *who
a caller is* has to happen before a tenant is known, and the policy refuses an
unbound read — so the two requirements are in direct conflict. Each function
answers "which tenant does this identifier belong to" and **returns no row**.
Callers resolve a tenant, bind it, then read through the ordinary scoped path.

**Migrations moved off the runtime connection** to a separate owner credential
used at boot, which grants the runtime role and closes. The runtime role never
holds schema rights.

## The property, stated accurately

The exemption discloses the **shape** of the deployment — which tenants exist,
and which tenant a given user, credential, room, bridge or connector belongs to
— and **no row of any tenant-scoped table**. A boundary on data, not on
metadata.

An earlier draft of this description claimed something stronger and it was
false three ways. Two are now closed. The third is not closable here: `users`
is global by design, so anyone able to call the lookups can reconstruct the
membership graph. That is recorded in the code, in the design doc, and pinned
by a test as a known property. **If cross-tenant metadata disclosure is not
acceptable, that is a Phase 2 decision about scoping users**, not a bug here.

## What a reviewer should attack

**The seven functions.** They are the whole exemption surface. If one of them
can return a row, or be called with something the caller does not already hold,
the design is wrong rather than buggy.

**The drift guard.** Binding a tenant around an already-open transaction used
to be a silent no-op — the database is told once, at transaction start. It now
raises. Two hooks cover reads and writes; a flush bypasses the first, which was
a real gap found by removing the hook and watching nothing fail.

**The deployment defaults.** A managed Helm deployment now has a role created
for it by a pre-upgrade hook, and `DB_USER` moves off the superuser. `mode:
existing` renders byte-identically to before.

## Also in here

Two pre-existing cross-tenant bugs found by running two tenants for real: the
default-bridge lookup that 500s for *both* tenants once a second one exists,
and an invite mechanism keyed on an identifier every tenant's admin client
shares, which silently dropped room memberships. Plus the per-tenant lookup
sweep (#429) and the last of four context-token instances (#430), folded in.

## Verified

2941 passing. The service was stood up under the restricted role against a real
Postgres: boots, migrates as the owner, grants, self-checks, both front doors
authenticate, a message crosses between agents — then a second tenant, each
seeing only its own rooms and agents, 404 both directions across the boundary.
Populated tables throughout, which matters: the policy is only evaluated when a
scan yields rows, so an empty table proves nothing.


petr-sandbox:
A few things to consider:

1. The integration suite is broken by this PR — 11 errors, all at fixture setup.
tests/integration/conftest.py:411 → user_store.create → require_tenant_id() → TenantNotBoundError. Every integration test errors before it runs. It builds its own session factory (:355) and binds nothing, so it's the one suite running production-shaped — and just test deselects integration (-m 'not integration'), so the suite is green and this is invisible. On the base branch the same line succeeds. The PR touched that file by 2 lines (adding TenantStore) and stopped one line short.

2. You can gut the migration and the whole suite stays green.
I changed 265ed188ad6f three ways at once — WITH CHECK (true), drop the empty-string check, and make _enable_rls_ddl return SELECT 1 so no table gets RLS enabled at all — and tests/switch_core/db/ stayed at 270 passed. test_migration_parity too: compare_metadata doesn't look at policies or relrowsecurity. The only thing pinned is the frozen table list (test_row_level_security.py:409), not a byte of the DDL. So an Alembic-built production database with zero row-level security passes everything — which is precisely what that test's own docstring says it exists to prevent.

The fix is cheap and I verified it: run the chain into one scratch DB, create_all into another, diff pg_policies + pg_class.relrowsecurity + pg_proc.prosrc. Today they're identical — 38 policies each, same quals, same function hash — so the test passes on day one and starts earning its keep on day two.

3. tenant_scope around an already-open session is a silent no-op on the database side.
Reproduced twice independently:

[1] before any bind:              None
[2] inside tenant_scope, same tx: None   ← binding had no effect
[3] after commit + tenant_scope:  '000...000'




after_begin fires once, at transaction begin; tenant_scope only rebinds the contextvar. main.py:565 is exactly this shape — and its comment says the row "names tenant zero itself rather than inheriting a binding there isn't one of". It names nothing. The row gets its tenant from the column default, which is the accident the comment claims to be avoiding. Today it fails open because the owner bypasses; under a restricted role the insert is refused. Worth fixing here since it's the one place the design doc points at as its known gap. The other 18 tenant_scope uses wrap a fresh session and are fine.

petr-sandbox (approved):
Looks good.

PR #427, opened 2026-09-10
Add persistent SDK sessions, native context controls and durable attachments

raullopez-sandbox:
Switch Console uses persistent SDK hosts for local and SSH sessions. Closing Console leaves execution running. The server owns authorization, tenant isolation, leases, recovery epochs, command receipts and answer arbitration. Tmux agent execution is removed.

Reset retains transcript history and starts a fresh native conversation. An interrupted reset or a missing Codex rollout requires an explicit fresh-reset decision; Switch never chooses a replacement conversation or repeats uncertain actions automatically. Recovery persists uploads, acknowledgements and room cursor changes. Process fencing requires confirmed provider-group exit before another host starts. Discovery restores room bindings and isolates failed sessions.

The composer provides searchable model and effort menus plus Context actions. Confirmed model choices persist across restart. Native compaction is exposed only where supported. Picker, paste and drop attachments use authenticated durable references; the execution host downloads and verifies the actual bytes. Skills, MCP, credentials and native configuration load on the execution machine. Discord sender-name rejection uses a safe, stable name with visible agent attribution.

| Provider | Reset | Model selection | Native compaction | Attachment delivery |
| --- | --- | --- | --- | --- |
| Claude Code | Yes | Catalog and effort | When advertised | Images as bytes; files staged |
| Codex | Yes | Catalog and effort | App-server compaction | Images and staged file mentions |
| OpenCode | Yes | Connected models and variants | Session compaction | Staged file URLs; model MIME support checked |
| Gemini CLI | Yes | ACP catalog | Unavailable | Image/resource bytes |
| Cursor | Yes | ACP catalog | Unavailable | Image bytes and staged file references |

Validation: final provider suite 240 passed; runtime suite 113 passed; Discord/display-name suite 177 passed. Ten recovery tests passed against real PostgreSQL. Provider/runtime types, Console types, affected lint and Discord adapter mypy passed. Earlier isolated local SSH checks covered bundle startup, staging, retries, native skill paths and MCP environment with simulated provider/download inputs.

Live Discord and Console checks verified both channel-creation directions, chat-triggered Claude and Codex sessions, sidebar room binding, transcript replay, unaddressed-message cues, queued delivery during approval, Codex interrupt/reset, and explicit recovery of preserved Claude and Codex failures. A later Codex process crash resumed the same native conversation ID. After fixing the server-reset cursor regression, the original waiting Discord message arrived without being resent. Reserved-name delivery also passed in native Discord with visible attribution. Approvals were answered through Console.

Earlier live provider checks covered context controls, file contents, project skills and MCP with Claude, Codex, OpenCode and Gemini; image recognition passed with Claude, Codex and Gemini. The tested OpenCode model rejects images. Cursor generation remains blocked by its account limit. The current Discord run covers Claude and Codex only.

Remaining limits: the in-memory room stream can detect a server restart only while the old cursor is ahead of the new buffer; it has no persistent generation identity. Legacy delivery acknowledgements without substitution records require server-evidence reconciliation. Discord-only approval cards and publication changes remain on the separate messaging companion branch, which is not merged here. Automatic attachment retention is not implemented. No authenticated external remote host was supplied, so local SSH tests do not establish external deployment readiness.


raullopez-sandbox:
Claude permission settings: c7af58c3 lets the CLI resolve its native permission mode when bypass is off and removes the forced approval hook. The permission callback remains available for requests that need a human.

Validation: 45 Claude adapter tests pass; typecheck, lint and formatting pass. A live SDK/CLI check inherits `acceptEdits` from local settings. The installed CLI resolves `auto` to `default`; automatic mode is not verified. This uses the installed SDK 0.3.260 native-resolution option, which is not declared in its public Options type. Recheck it on SDK upgrades. Console UI verification is still pending.

raullopez-sandbox:
Computer Use verification of c7af58c3: created a fresh Claude session in the development Console with Bypass permissions off and local `permissions.defaultMode: "auto"`. A Write request produced an approval card. Allowed it once; the file contained the expected marker and the transcript showed completion. Stopped the test session and removed the temporary settings.

The new Claude process had no `--permission-mode` argument. The adapter no longer forces default, but this run does not prove auto mode is active: the standalone SDK initialization check reported default with the same setting. The exact auto behavior remains unresolved; do not count this as a passing auto-mode test. Screenshot captured through Computer Use.

raullopez-sandbox:
Transcript scrolling fixed in 1bb5daf7. The view follows content and viewport size changes while at the bottom. Scrolling more than 24 pixels away disables following; returning to the bottom enables it again. This also covers streaming text and expanding request cards.

Verified through Computer Use in the development Console with a live Claude session: a 45-line response ended in view, then a follow-up message and reply both remained visible without manual scrolling. Screenshot captured. Typecheck, lint and diff checks passed. The scroll-up preservation path was reviewed in code but not completed in the native UI test because the automation rejected the paging keys.

raullopez-sandbox:
Live model-switch verification through Computer Use on 1bb5daf7:

| Provider | Switch | Result |
| --- | --- | --- |
| Claude | Sonnet → Opus | Applied; reply completed |
| Claude | Opus → Sonnet | Applied; reply completed |
| Claude | Sonnet → Fable | Applied; reply completed |
| Codex | Sol → Astra | Applied; reply completed |
| Codex | Astra → Sol | Applied; reply completed; recalled the prior Astra verification marker |

Each provider used one existing conversation throughout its sequence, with no host reset between changes. All sessions returned to ready. The reported unrecoverable-session failure did not reproduce in these runs. These checks cover idle sessions between turns, not an in-flight turn or a transport failure during a switch. The known error path that faults the host after a model-change exception remains to be addressed; these successful cases do not validate that failure path.

raullopez-sandbox:
Transcript UI fixes in 09442263:

- Model-change notices keep the message anchor where they arrived, so later turns appear after them.
- Approval and question cards stay with their turn. Settled requests collapse to a result instead of keeping disabled action buttons visible.
- Sending, queued and waiting states give feedback before the first provider output.
- Rename “Restart host” to “Restart session process.” Local SDK sessions also run in a persistent background process.

Validation: 31 shared-package tests pass, including a notice-position regression; Console typecheck and focused lint pass. Computer Use confirmed a model-change notice remains before a new completed turn, queued feedback is visible immediately after sending, and historical approvals are collapsed beside their original turns. Screenshots captured.

Existing limitation: snapshot responses do not include historical notices; this change anchors notices received by the live client, but does not add notice history to snapshots.

raullopez-sandbox:
Compared with T3 Code source at 6f00d3881a197dd33c2cb43c6a11a9e759e56089. This was source and test inspection, not a live T3 run.

| Behavior | Comparison |
| --- | --- |
| Follow streamed content at the bottom | Same basic behavior. Both stop following when the user scrolls away and resume at the end. |
| Send while reading history | Different. T3 calls scrollToEnd for subsequent sends. Switch follows only if already at the bottom. T3 also frames the first message near the top. |
| Feedback before response | Same purpose. T3 has Working/Thinking timeline rows; Switch has Sending/Queued/Waiting. T3 also has optimistic user messages. |
| Pending approvals | Different placement. T3 uses a composer panel; Switch uses inline turn cards. Both remove actionable controls after settlement; Switch keeps a collapsed historical result. |
| Notice chronology | Switch now anchors live notices before later messages. T3 orders timeline entries by createdAt with tie-breaking. No equivalent literal model-changed notice was found, so exact parity is not established. Switch snapshots still omit notice history. |
| Restart control | Switch-specific persistent-session control; no matching always-visible control was found in T3 ChatView. |

Reference files: T3 apps/web/src/components/ChatView.tsx (send-time scrolling and free-scroll behavior), chat/timelineScrollAnchoring.ts and tests, chat/MessagesTimeline.logic.ts (working/thinking), chat/ComposerPendingApprovalPanel.tsx, packages/client-runtime/src/pendingRequests.ts, apps/web/src/session-logic.ts (timeline ordering).

The reported bottom-follow, waiting-feedback and settled-card issues are addressed, but full T3 parity should not be claimed. No code changes were made for this comparison.

raullopez-sandbox:
Fresh live approval regression verified through Computer Use on 09442263: requested `sleep 2` in a Codex session, approved once, and observed completion. The approval buttons disappeared and the result collapsed to “Approved · sleep 2”. A subsequent prompt and completed reply appeared below the sleep turn; its approval result stayed in the original turn. Screenshots captured. No further code changes needed.

raullopez-sandbox:
Fixed the persistent sidebar spinner in 0859ddc1. The sidebar previously relied on hook-driven activity while the transcript used SDK state. SDK snapshots and session events now update the stored activity and notify the renderer; discovery also synchronizes existing background sessions. Ready/stopped clears working, pending requests show awaiting-input, and errors remain visible. A connection loss does not imply execution stopped.

Validation: three focused tests, Console typecheck and lint pass. Computer Use with a live 100-line Codex response showed the sidebar spinner during running and no spinner after ready, with the process still connected. Screenshots captured for both states. Background discovery is code-covered by the same sync path; the live test used an open conversation.

raullopez-sandbox:
Stop and Resume implemented in 3ee81f35. Stop ends execution and retains the conversation. The stopped view now shows a Resume session button and explains that interrupted work will not be repeated.

Explicit restart/resume records a one-use operation in the host journal after process fencing. It clears the stopped state and uses the existing native conversation and server-issued recovery epoch. Ordinary reconnect leaves a stopped session stopped. A consumed resume operation cannot undo a later Stop. Existing unknown command outcomes and interrupted turns are not resent; unavailable native conversations still require an explicit recovery decision.

Validation: 241 provider tests pass, including stop/reconnect and resume/stop/same-operation-restart regressions; Console typecheck and lint pass. Live Computer Use against the local PostgreSQL-backed setup: stopped a Codex session, saw Resume, resumed it, and received a new reply that correctly recalled the earlier conversation marker. Screenshots captured before and after. Other providers share the host path but were not separately live-tested for this change. No external SSH verification or archive redesign was performed.

raullopez-sandbox:
Provider readiness and Gemini model selection (3a45c8d0)

Add Agent now checks the provider on the execution machine. The shared worker checks again with the session environment before starting execution. A confirmed missing login blocks startup and gives the native sign-in command. Unknown status remains a visible warning in setup. Gemini now has a model dropdown populated by its native ACP model list.

Compared with T3 at 6f00d388: use native status data and keep unknown distinct from unauthenticated. Codex uses account/read; Cursor uses about; OpenCode checks connected backends. Claude uses auth status (T3 also uses SDK initialization data). T3 does not list Gemini CLI, so its ACP check is specific to Switch.

Validation:
- 253 provider tests and 19 model-catalogue tests passed; provider/Console typecheck and lint passed.
- Live Computer Use: Claude, Codex and Cursor displayed signed-in status; Gemini returned models; OpenCode reported connected backends. Selected Gemini 2.5 Pro from the final dropdown. Screenshots captured in the work session.
- Real isolated-home checks: Claude and Codex returned unauthenticated. Gemini still found usable credentials, so that is not a signed-out test. Protocol tests cover Gemini missing-key rejection, transport failures, and cleanup.

Limits: this verifies existing native logins, not five new interactive browser sign-ins. Sign-in still runs in the provider CLI, followed by Check again. No external SSH host was tested in this change. Status does not prove model entitlement or quota. Gemini's metadata probe creates an empty native session without sending a prompt.

raullopez-sandbox:
Unified model fields (8861425e)

All providers now use one editable model picker in agent configuration. Choose a model from the execution host or enter an ID. Blank uses the provider default. Loading and failed catalogue reads are visible. Gemini no longer becomes a fixed Default-only select when its catalogue is unavailable.

Claude, Codex, Cursor and Gemini model lists come from native SDK initialization, with no prompt sent. OpenCode keeps its existing host catalogue. Model probes stop their provider session after reading the list and have a time limit. Escape preserves a custom model ID. Bare model IDs no longer appear twice as a heading and an option.

Checks: 25 catalogue/grouping tests and 8 real-browser tests passed, including Escape preserving text. Console and provider typecheck/lint passed. Live native probes returned 5 Claude, 6 Codex, 38 Cursor and 6 Gemini choices. Computer Use verified Claude and Gemini lists through the same control and selected Claude Fable. A screenshot captures the Gemini list. Other providers' native lists were verified programmatically; no external SSH run in this change.

Rebuild the Console packages and host bundle, then restart Console to load the new main-process model discovery.

raullopez-sandbox:
Room control follow-ups restored in 901d3f35.

After a confirmed `!reset` or `!compact`, the server queues one durable prompt to reconnect, read room context, restore the previous role, and confirm completion in the original thread. The role is captured before execution. Failed or unknown controls do not announce success. Recovery uses the current epoch and does not replay an uncertain follow-up.

Validation: 17 room tests and 69 other session, tool-contract, and migration checks passed on PostgreSQL. Ruff and mypy passed. All five connector skills describe the same behavior.

Live verification: a Discord command ran native Codex compaction and produced a targeted success reply in its original thread. Database receipts confirmed both the compaction and follow-up. Computer Use later submitted an unintended compact-all draft in the disposable test room while returning stale controls; both test sessions completed compaction and remained ready. UI input was stopped. Live reset and live role reacquisition remain unverified; both paths have PostgreSQL regression coverage. No external SSH verification was performed for this change.

raullopez-sandbox:
Warning copy updated in cee01b0f.

The automatic-session failure now explains its effect and what the user can do. Raw errors are collapsed under Technical details. The section is named Session availability, and Room watcher is now Automatic sessions. This changes the presentation; it does not repair the listener after a delivery gap.

Verified the updated agent page in the running Console. Formatting, typecheck, and targeted lint passed. The linter's native worker crashed on the first run; the single-worker retry passed. The live page showed the available state; the unavailable-state wording was checked in source.

PR #428, opened 2026-09-10
fix(logging): never restore a log context token in a context that did not make it

wojtyniakAQ:
## Summary

`unbind_log_context` calls `Token.reset`, which refuses to cross contexts. `log_context`'s cleanup runs that reset from a plain `finally`, which assumes the block is always *resumed* to its exit. It isn't always: a coroutine suspended inside `log_context` can instead be dropped and later *finalised* by the garbage collector closing it, and the collector runs that `finally` in whatever context it happens to be in — not the one that created the token. `Token.reset` then raises `ValueError: ... was created in a different Context`.

This is latent rather than live today: it needs a frame that is finalised by the collector instead of resumed normally, and nothing currently drops a coroutine mid-`log_context` block in a way that reaches that path. That's a statement about current allocation behavior, not about the code — it was found while chasing a related bug elsewhere where the identical pattern turned an occasional test flake into a 100% reproducible failure once unrelated changes shifted allocation timing enough to move a GC pass inside the test window. Leaving it means the next such shift turns this one live too, at a time with no connection to whatever change actually triggers it.

Fix: skip the restore when the block is unwound by `GeneratorExit` instead of resumed. There's nothing to restore into anyway — the context the token belongs to is unreachable, which is why the frame is being finalised at all.

This now covers all instances of the bug, not just the context-manager form. The five raw `bind_log_context`/`unbind_log_context` call sites noted below as outstanding are done: `request_context.py` and `bridges/agent/auth.py` now use the `log_context` context manager directly; `operations/callctx.py` gained its own `call_context` context manager (it restores a call-context token alongside the log token, so it needed the same protection under one name), and `mcp/server.py` and `api/operations.py` — the two callers that held `set_call_context`/`reset_call_context` in a `try`/`finally` — now use it. The shared finalisation check moved into a reusable `restore_unless_finalising` helper in `logging_context.py` so it is written once rather than duplicated per token type.

`bind_log_context`/`unbind_log_context` stay public: `gateway/auth.py` binds one deliberately with no matching unbind (left for the outer request middleware to discard), and `callctx.py` composes a log token with its own call token, so both still have a legitimate caller outside the module.

Refs CHOO-2698.

## Test plan
- [x] Added a regression test that drives a coroutine into a suspended `log_context` block in one context and closes it from another, reproducing the `ValueError` on the old code and passing on the new one.
- [x] Added the same regression coverage for each of the five call sites (`request_context.py`, `bridges/agent/auth.py`, `operations/callctx.py`'s `call_context`, `mcp/server.py`'s `CallContextMiddleware`, and `api/operations.py`'s `call_operation`), including one exercised through the real HTTP operations front door rather than the helper alone.
- [x] `just test`
- [x] `just check`
- [x] `just typecheck`

### DIFF
--- a/core/switch_core/bridges/agent/api/operations.py
+++ b/core/switch_core/bridges/agent/api/operations.py
@@ -26,11 +26,7 @@ from fastapi import APIRouter, Depends, Header, HTTPException
 from switch_core.bridges.agent.auth import get_agent_from_scope
 from switch_core.bridges.agent.dependencies import get_protocol
 from switch_core.bridges.agent.operations import all_operations, get_operation
-from switch_core.bridges.agent.operations.callctx import (
-    CallContext,
-    reset_call_context,
-    set_call_context,
-)
+from switch_core.bridges.agent.operations.callctx import CallContext, call_context
 from switch_core.bridges.agent.protocol.connections import UnknownConnectionError
 from switch_core.bridges.agent.protocol.service import ProtocolService
 from switch_core.db.models import Agent
@@ -103,14 +99,11 @@ async def call_operation(
     if missing:
         raise BadArgumentsError(f"{operation} requires: {', '.join(missing)}")
 
-    token = set_call_context(CallContext(agent_id=agent_id, session_key=connection_id))
-    try:
+    with call_context(CallContext(agent_id=agent_id, session_key=connection_id)):
         result = fn(**call_args)
         if inspect.isawaitable(result):
             result = await result
         return result
-    finally:
-        reset_call_context(token)
 
 
 # ── HTTP router ──────────────────────────────────────────────────────────────
--- a/core/switch_core/bridges/agent/auth.py
+++ b/core/switch_core/bridges/agent/auth.py
@@ -20,7 +20,7 @@ from switch_core.db.tenant_lookup import (
     tenant_of_agent_oauth_client,
     tenant_of_api_key,
 )
-from switch_core.logging_context import bind_log_context, unbind_log_context
+from switch_core.logging_context import log_context
 from switch_core.tenant_context import tenant_scope
 
 logger = logging.getLogger(__name__)
@@ -162,12 +162,10 @@ class BearerAuthMiddleware:
             # path resolves straight to an Agent with no ApiKey row, so it
             # falls back to the agent's own tenant.
             tenant_id = api_key.tenant_id if api_key is not None else agent.tenant_id
-            with tenant_scope(tenant_id):
-                log_token = bind_log_context(agent_id=agent.id, tenant_id=tenant_id)
-                try:
-                    await self.app(scope, receive, send)
-                finally:
-                    unbind_log_context(log_token)
+            with tenant_scope(tenant_id), log_context(
+                agent_id=agent.id, tenant_id=tenant_id
+            ):
+                await self.app(scope, receive, send)
             return
 
         # Registration token: pass through (handler validates again). MCP rejects.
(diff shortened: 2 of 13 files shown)


wojtyniakAQ:
Scope note, so this is not mistaken for closing CHOO-2698 entirely.

This fixes the context-manager form, which is the one that can be finalised by
the collector while suspended — that is the reachable case, and the
reproduction in the test drives exactly it.

**Five call sites still use the raw `bind` / `unbind` pair in a `try`/`finally`**
rather than the context manager: `request_context.py`, `bridges/agent/auth.py`,
`operations/callctx.py`, `mcp/server.py` and `api/operations.py`. They carry the
same exposure in principle. They are left alone here to keep this a one-file
change, not because they are safe.

Worth stating plainly why "in principle" is not the same as "ignore": the
sibling instance of this bug sat latent for months and then became a
100%-reproducible failure, purely because unrelated work changed allocation
patterns enough to move garbage collection inside the window. "Cannot happen
today" describes today's allocation behaviour, not the code.

So CHOO-2698 should stay open after this merges, pointed at those five.

**Update: the five are done.** `request_context.py` and `bridges/agent/auth.py`
now go through the `log_context` context manager directly. `operations/callctx.py`
gained a `call_context` context manager built on the same finalisation-safe
helper, and both of its callers — `mcp/server.py`'s `CallContextMiddleware` and
`api/operations.py`'s `call_operation` — use it instead of the raw
`set_call_context`/`reset_call_context` pair. Each site has its own regression
test reproducing the `ValueError` on the old code. CHOO-2698 can close with
this PR.

petr-sandbox:
Just some bits Claude suggested:

1. GeneratorExit doesn't mean "the token's context is gone"

restore_unless_finalising treats GeneratorExit as proof that restoring is impossible. It only proves the frame is being finalised. The GC runs that finalisation in whatever context happens to be current — which is sometimes the very context that created the token. There, Token.reset would have worked, and the PR now skips it, leaving the variable bound for the rest of that context.

This test passes on main and fails on the branch:

```
def test_finalising_in_the_creating_context_still_restores() -> None:
    async def body() -> None:
        with log_context(request_id="dropped"):
            await _Suspend()

    with log_context(request_id="caller"):
        coro = body()
        coro.send(None)   # entered in THIS context — token belongs here
        coro.close()      # collector happens to run here too
        assert current_log_context().request_id == "caller"
E   AssertionError: assert 'dropped' == 'caller'
```

The docstring's "There is also nothing left to restore, since the context the token belongs to is unreachable" is only true of the cross-context half.

Why it matters more than it looks: gateway/auth.py:96 binds user_id with no unbind of its own, relying entirely on RequestContextMiddleware's reset to discard it. Skip that reset in the same-context case and one user's id stays stamped on subsequent log lines. For a change whose entire purpose is making logs attributable, silently wrong attribution is a worse outcome than the loud ValueError it replaces — and it's the failure mode CLAUDE.md ranks last.

Suggested fix — attempt the restore, swallow only the cross-context ValueError, and only while finalising:

```
    finally:
        try:
            restore()
        except ValueError:
            if not finalising:
                raise
```

I applied that in the worktree: all 25 of the PR's tests plus the new one pass (26/26).

2. The tests only cover one half

All six regression tests are the same copy-pasted shape — contextvars.copy_context().run(coro.send, None) then coro.close(). Every one exercises the foreign-context case; nothing exercises finalisation in the creating context, which is why #1 got through. That's the test to add, and it belongs next to the existing one in test_logging_context.py.

3. Six copies of the same scaffolding

_Suspend and the drive-then-finalise helper are duplicated verbatim into all six test files. One shared helper (a conftest.py under core/tests/switch_core/ — there isn't one yet) instead. Related: five near-identical per-call-site tests for one shared helper is a lot; the unit test plus the one that goes through the real HTTP front door carry most of the value.

4. set_call_context / reset_call_context now have no production callers

After this PR the only users are four test files (test_reference_types_e2e, test_connect_to_room_eviction, test_operations, test_read_context_room_argument). The PR body's justification for keeping them public is about bind_log_context composition, which doesn't apply to these two. Those tests should switch to with call_context(...) so they exercise the path production actually takes, guard included.

petr-sandbox (approved):
Looks good.

amaudruz (approved):

PR #437, opened 2026-09-11
feat: platform sender kind for agent addressing (CHOO-2719)

abeldantas:
## Brief

Adds "platform" as a third sender kind so Switch can address agents as itself — not impersonating a user, not pretending to be an agent. An agent must explicitly opt in to receive platform messages; the default is deny, with no configuration needed. Room templates can now send a kickoff message after provisioning, posted as the Switch app through normal permission checks. Creator impersonation is still available but only when explicitly chosen, and it's visible (the marker carries the creator's name, not silent).

CHOO-2719: https://sandboxquantum.atlassian.net/browse/CHOO-2719

## Summary

- `SenderKind` extended to `"user" | "agent" | "platform"` in `addressing.py`
- New `platform: bool` field on `AddressingRule` — denied even by open policies (no rules), opt-in via `platform=True`
- `resolve_sender` recognises the admin client (`type="admin"`) as `SenderPrincipal("platform", ...)`
- `PLATFORM_MARKER` (`com.switch.platform`) content marker with optional `on_behalf_of` for visible creator impersonation
- `AdminClient.send_platform_message()` sends addressed messages as the Switch app
- `sender_kind` field on `MessagePayload` and `read_context` timeline entries
- Bridge routes platform-marked messages through the admin/app rendering path
- `RoomSpec.kickoff` — post-provision message, default `sender: "platform"`, opt-in `sender: "creator"`
- No DB migration — existing JSONB policies default `platform: false` via Pydantic

## Test plan

18 new unit tests covering the platform addressing model (deny-by-default for open and closed policies, room/group scoping, no fallthrough to agents dimension, backward compat for stored policies without the key, policy factories), plus resolve_sender admin-client detection, and kickoff spec parsing. Full suite green at 435 relevant tests, lint and typecheck clean.

### DIFF
--- a/core/switch_core/addressing.py
+++ b/core/switch_core/addressing.py
@@ -14,8 +14,12 @@ Model:
       * `agents`      — matched when the sender is another agent
   - Each dimension is either ``"*"`` (any), or an explicit list of ids. An
     empty list ``[]`` is the "none" value: it matches nothing. The sender is
-    exactly one kind (user XOR agent), so a rule that should admit only humans
-    sets ``agents: []`` and vice-versa.
+    exactly one kind (user, agent, or platform), so a rule that should admit
+    only humans sets ``agents: []`` and vice-versa.
+  - A rule carries a ``platform`` boolean (CHOO-2719). When True, the rule
+    admits the Switch platform as a sender. Platform is denied by default —
+    even an open policy (no rules) refuses it — so an agent must explicitly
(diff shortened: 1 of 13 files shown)


abeldantas:
Folded into #405 and merged with it as 2e0706f5: the platform sender kind, deny-by-default with a per-rule opt-in, and the kickoff posted by Switch with the creator's authority. Nothing here is left to merge separately.

PR #438, opened 2026-09-11
docs(multi-tenancy): land the design and phased plan on main (CHOO-2729)

wojtyniakAQ:
## Summary

- Adds `docs/old/multi-tenancy.md` — the multi-tenancy design and the phased plan the work has been following. It was written on the spike branch and its pull request was closed unmerged, so it has never been on `main` even though all five phase tickets cite it as the source of truth, and Phase 1 shipped against it.
- Corrects the status header. It said "design spike — not implemented", which stopped being true when Phases 0 and 1 landed. It now names which phases exist and says plainly that the code wins where the two disagree.
- Indexes it in `docs/README.md` and `CLAUDE.md`, and while there indexes `docs/old/multi-tenancy-phase1-db.md` and `docs/old/rds-migration.md`, which are already on `main` but appear in neither list.

No content changes beyond the status header — the design itself is the record of what was intended, and editing it after the fact would destroy that.

## Test plan

- [x] Documentation only; no code touched, so no lint, typecheck or test run.
- [x] Read for public-repo safety: no credentials, internal hostnames, IPs, account names or personal data. Ticket keys only, which the repository's own guidance permits in design notes.
- [x] Every path added to the two indexes exists in the tree.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -238,6 +238,14 @@ Tests live in `core/tests/switch_core/` mirroring the module structure. Uses pyt
 - `docs/old/LOCAL_DEVELOPMENT.md` — running Switch locally for development:
   `just` recipes, which port serves what, connecting Switch Console to a
   local server
+- `docs/old/multi-tenancy.md` — why Switch is multi-tenant the way it is: the
+  tenant model, sign-in and onboarding, one official messaging app per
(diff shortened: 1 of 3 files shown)


petr-sandbox (approved):
Very nice.

PR #439, opened 2026-09-11
docs(multi-tenancy): design several workspaces per person (CHOO-2723)

wojtyniakAQ:
## Summary

Design only — no code. Covers `CHOO-2723`, `CHOO-2722`, `CHOO-2724` and `CHOO-2726` as one document, because they are four questions about the same object and answering them apart produces answers that do not fit together.

Revised after an adversarial review, which overturned one decision and found three false claims — see the second commit's message.

## ⚠️ Conditions attached to merging and deploying this

Two consequences of the design that must be true before any of it reaches a deployment. They are recorded here so they are known at review, and again at merge, rather than discovered afterwards.

**1. Removing a member must revoke their credentials in the same transaction — or member removal must not ship.**
API keys and agents resolve their workspace from their own row and never consult membership, so removing someone leaves every key they minted still working. Worse, those keys become invisible: the key list runs on the tenant-bound request session, so a key in a workspace you have left cannot be seen or revoked by you, and its owner is no longer a member for anyone else to find it under. A removal that leaves working credentials behind is worse than no removal at all, because it looks like it worked. **Do not merge a member-removal route without the cascade.**

**2. The multi-workspace 409 is a breaking change for every client at once.**
Anyone holding two memberships starts getting a "choose a workspace" response from every endpoint. Console's workspace switching is explicitly not in this phase. **It ships behind a flag, or after a Console release that can answer it — not before.**

## The decisions

- **Split the two roles, and sequence that first.** *(Confirmed by wojciech.wojtyniak; filed as its own ticket.)* The role that authorises anything today is the global one on the user; the per-tenant role is written and never read. So "the creator of a workspace owns it" grants nothing, and any bootstrap that sets the global bit makes that person an administrator of every workspace they later join. The global bit narrows to a **deployment operator**; the per-tenant role becomes the input to tenant-scoped authorisation.
- **A request's tenant is selected by a claim in the session cookie and authorised by the database on every request.** True of the browser session and *false of bearer credentials* — hence condition 1 above.
- **Listing a person's workspaces runs before any request session is open.** The first version looped bound sessions inside the handler; the request session already holds its connection and the suite pins one per request, so that needed a second — as would the exempt lookup the first version rejected on principle.
- **Two new `SECURITY DEFINER` lookups, and only two.** Resolving an invitation token and a claimed email domain both happen before a tenant is bound, which is the shape the existing seven exist for. A lookup returning workspace *names and roles* is still refused: that answers *what row*, not *which tenant*.
- **The administrator bootstrap needs no new mechanism.** Identity linking is not refused for a verified address — the earlier account of this was wrong — so an operator who points the configured administrator address at one they control at the provider lands in the administrator account on sign-in. Corrected on the setup page by #440.

## Test plan

- [x] Documentation only; no code touched.
- [x] Every claim about current behaviour re-checked against `main` after the review, including the ones the review corrected: the global role bypass in `authz.py` and `require_admin`, the verified-email linking branch in `user_store.py`, the bearer path's tenant resolution in `bridges/agent/auth.py`, the pinned one-connection-per-request test, and the absence of any `email_verified` column.
- [x] No credentials, internal hostnames or personal data.

### DIFF
--- a/docs/old/multi-tenancy-phase2-tenants.md
+++ b/docs/old/multi-tenancy-phase2-tenants.md
@@ -1,0 +1,201 @@
+# Phase 2: several workspaces per person
+
+Status: design, not built. Covers `CHOO-2723` (several memberships),
+`CHOO-2722` (the tenant API and invitations), `CHOO-2724` (joining by email
+domain) and `CHOO-2726` (how anyone becomes an administrator). They are four
+tickets and one design: each is a different question about the same object,
+and answering them separately produces four answers that do not fit together.
+
+Phase 1 built the boundary. This phase is about crossing it deliberately —
+letting one person belong to several tenants, and giving them a way to create,
+join and choose between them. Read `multi-tenancy-phase1-db.md` first; this
+document assumes its vocabulary and does not repeat it.
+
+## 1. What exists, exactly
+
+- `tenants` — `id`, `slug`, `name`, `created_at`. Nothing else.
+- `tenant_members` — `(tenant_id, user_id)` as the primary key, plus a `role`
+  constrained to `owner` / `admin` / `member`. The roles are recorded and mean
+  nothing yet.
+- One tenant exists in every deployment: tenant zero, seeded by migration with
+  slug `default`. Sign-in binds it explicitly rather than by fallback.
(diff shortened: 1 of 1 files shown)


petr-sandbox (approved):
Makes sense.

PR #441, opened 2026-09-11
ci(migrations): re-check open PRs when a new head lands on main (CHOO-2689)

wojtyniakAQ:
## Summary

Two migrations chained off the same parent are each green in isolation and produce a second head only once both merge — at which point the server crash-loops on boot, because it self-migrates with `alembic upgrade head` and that is ambiguous with more than one. It happened twice in 48 hours (#404 with #397, then the templates migration) and was hand-fixed with merge revisions both times.

**The gap is not where the ticket says.** A manual `alembic stamp` cures nothing permanently, and the per-PR check is not testing the wrong tree: `actions/checkout` already checks out GitHub's merge preview on a `pull_request` event, verified against a real run's logs. The gap is staleness — GitHub recomputes that preview as the base moves but nothing re-runs the workflow, and merging is not gated on the branch being current, so a day-old green check still satisfies the merge button.

So: when a migration lands on `main`, re-check every *other* open PR that also adds one, and post a failing status on it immediately rather than letting the next deploy find out.

**It reads the chain statically, and that is the security-relevant part of the design.** The obvious implementation overlays the other PR's files and runs the existing test — but that test builds an Alembic `ScriptDirectory`, and Alembic **imports every revision module** to construct the graph. On a `main` workflow run holding a token that can write statuses, in a public repository, that is executing unreviewed Python from any pull request anyone cares to open. This parses `revision` and `down_revision` out of the file text with a regex and `ast.literal_eval` instead — the same way the existing `test_revision_ids_are_unique` already reads revision ids, and for the same reason — and derives heads as "any revision that is nobody's parent". Nothing is imported, nothing is checked out, and the job needs neither a database nor the project's dependencies.

The per-PR run keeps using the real test unchanged: there, the code being imported is the code under review.

## Test plan

- [x] Verified offline against the real 65-revision tree: healthy chain reports no problem.
- [x] Fabricated a sibling migration off the current head's parent → `expected exactly one migration head, got ['a3f61c02d5be', 'fake000000aa']`.
- [x] Fabricated a duplicate revision id and a dangling parent → each reported.
- [x] Merge revisions parse correctly, including the tuple form of `down_revision` (checked against the real `b47e0c39a1f5`).
- [x] `just check`, `just typecheck` and `mypy` on the script directly all pass; workflow YAML validated.
- [ ] End to end needs a real collision: open two PRs each adding a migration off the current head, merge one, and watch the status appear on the other.

## Note for whoever has repository admin

Out of scope here, but found while investigating: the branch ruleset's required status checks are only "Detect changed trees" and "Secret scan". `Backend (core)` is not required, so a red test suite does not block a merge today.


petr-sandbox:
A few comments:

1. The union double-counts every shared migration — blocker

fetch_pr_revisions does git ls-tree -r FETCH_HEAD -- versions/, which returns the PR head's whole versions directory, not just the files it adds. check_pr then does base_revisions + fetch_pr_revisions(...), so all 65 migrations the two sides share appear twice.

```
base alone            -> ''                 (clean)
base + PR-head union  -> duplicate migration revision ids:
   {'04f27f37e474': ['04f27f37e474_oidc_identities_table.py',
                     '04f27f37e474_oidc_identities_table.py'], ...}
```

So every PR that touches migrations gets state: failure with "Merging this PR now would break the migration chain — rebase onto the latest main", unconditionally, and the job exits 1 on every main push. Worse, describe_chain_problem returns on duplicates first, so the two-head case it exists to catch is never reached.

The docstring's reasoning is right ("the versions directory only ever gains files, so a plain union ... is the merge result") — it's a set union that was implemented as list concatenation. Build the union keyed by filename, PR side winning:

merged = {r.filename: r for r in base_revisions}
merged.update({r.filename: r for r in fetch_pr_revisions(int(number))})
problem = describe_chain_problem(list(merged.values()))

With that, against the real tree: a PR chaining off the current head reports clean, and a sibling off the head's parent reports expected exactly one migration head, got ['a3f61c02d5be', 'sibrev00001']. That's the intended behaviour, and it's what the unchecked end-to-end box in the test plan would have caught.

2. gh pr list silently caps at 30

No --limit. The repo has 33 open PRs today:

```
gh pr list --state open --json number            -> 30
gh pr list --state open --limit 200 --json number -> 33
```

Three PRs get no check and nothing says so. Pass --limit 1000, or better --search to filter server-side.

3. The concurrency group cancels the run that matters

```
concurrency:
  group: pr-ci-${{ github.ref }}
  cancel-in-progress: true
```

On push, github.ref is refs/heads/main for every merge. Merge a migration PR, then merge anything else a minute later: the second run cancels the first, and because paths-filter on push only sees the second commit's files, migrations is false and the job is skipped. No statuses are posted, silently — on precisely the busy day this is meant to cover. Either give push runs their own group (pr-ci-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}) or set cancel-in-progress: ${{ github.event_name == 'pull_request' }}.

4. One bad PR aborts the whole sweep

run() is check=True everywhere and check_pr has no error handling, so a single gh/git failure — a transient API error, a fetch that can't resolve — raises out of the list comprehension in main(). PRs already checked keep their status; the rest get nothing, and the only signal is a red job on main that looks identical to a genuine finding. Wrap each PR, post an error status for the one that failed, and keep going.

That matters because two inputs can raise on ordinary content:

- Multi-line down_revision tuple. The regex captures to end of line, so down_revision = (\n "bb",\n "cc",\n) gives ast.literal_eval("(") → SyntaxError: '(' was never closed. Today's merge revisions fit on one line, so this is latent — but a long merge revision or a formatter change makes it live.
- A non-revision .py under versions/ → KeyError: 'revision' from values["revision"].

Both should be reported as a problem with that file, not a stack trace.

petr-sandbox:
All four addressed in 0f564fa8, and each verified against the real 65-revision tree the way the review was.

**1. The union — blocker.** Keyed by filename now, PR side winning, in a `merge_preview` helper that says why. Reproducing your numbers on the same tree:

```
base alone                    -> ''
OLD concat (base + prhead)    -> duplicate migration revision ids: {'04f27f37e474': [...
NEW union (base + prhead)     -> ''
PR chains off the head        -> ''
PR siblings the head          -> expected exactly one migration head, got ['a3f61c02d5be', 'sibrev00001']
```

You were right that the docstring's reasoning was sound and the implementation wasn't; the reasoning now lives on the helper that actually does the union.

**2. `gh pr list` cap.** `--limit 1000`, and it raises rather than proceeding if the listing ever comes back at the limit — silently checking a subset is the failure mode worth being loud about. Kept `gh pr diff --name-only` for the per-PR filter rather than moving to `gh pr list --json files`: that field is capped at 100 files per PR, and #444 has 217, so it would have swapped one silent truncation for another.

**3. Concurrency.** Push runs now carry the commit in their group:

```yaml
group: pr-ci-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}
```

Your first suggestion rather than the second: with `cancel-in-progress: false` on pushes the runs queue instead, and a third merge cancels the *pending* second one, which is the same hole a lane deeper.

**4. The sweep, and the two parse inputs.** Each PR now runs inside `check_pr_isolated`, which posts an `error` status naming the PR and carries on (and survives that post failing too). For the inputs:

- Multi-line `down_revision` — I swapped the regex for `ast.parse` and read the two assignments off the tree instead of the text. Still nothing imported and nothing executed, which was the whole reason for not using Alembic here, but an assignment is no longer assumed to be a line, so the multi-line tuple stops being a latent bug rather than a patched one. Confirmed against the real `b47e0c39a1f5` and a fabricated multi-line tuple.
- Non-migration `.py` — `RevisionProblem`, reported as a finding against that file. Same for a file that won't parse and for a non-string `revision`:

```
helper.py -> no module-level `revision` assignment -- every file under versions/ must be a migration
broken.py -> is not valid Python: invalid syntax (broken.py, line 1)
weird.py  -> revision is not a string literal: 1
```

`main` treats an unreadable file on *main itself* separately: it exits 1 saying so, rather than painting every open PR red for a problem none of them caused.

`ruff format`, `ruff check` and `mypy` all clean on the script; workflow YAML validated.

**On the Semgrep finding** — the taint it drew (`GITHUB_REPOSITORY` into an argument list) was removed in 88832913, and the rule reports nothing locally: 0 findings for `dangerous-subprocess-use-tainted-env-args` specifically and 0 for `p/python` + `p/security-audit` over the file. But the platform scan was still red on 540a9947, which it scanned *after* those commits landed — so either it's an unresolved finding that needs triaging on their side, or their Pro engine's interprocedural taint sees something the community rule doesn't. Watching what this push reports before deciding between a code change and an `/fp`.

Not addressed, and I don't think it should be here: the branch ruleset gap you found in the description (`Backend (core)` not being a required check) still needs someone with repository admin.

petr-sandbox:
**Parking this until Phase 2 (CHOO-2624) is done.** Not because of the review — all four points are fixed in 0f564fa8 and CI is green apart from the stale Semgrep finding.

The reason is scope: this touches `pr-ci.yml`, which is in the `shared` paths filter, and it changes the `concurrency` group for *every* run in the repo. Giving push runs their own group is the right fix for the hole in point 3, but it also means two merges in quick succession no longer cancel each other's 12-minute `Backend (core)` job — they both run. That is a repo-wide CI cost and behaviour change, and it wants the team's agreement rather than one reviewer's.

Nothing in the Phase 2 stack (#444 → #445 → #446 → #447, plus #442) depends on this. Until it lands, a second head that slips past a stale green check is still caught the way it has been: by hand, with a merge revision.

Two things to pick up when it comes off the shelf:

1. The Semgrep status is still red on the original finding from 06598b33 — no new comment, and the code that produced the taint went in 88832913. Needs an `/fp` or a console triage, not a code change (see the previous comment).
2. Worth considering splitting `migration-heads-on-merge` into its own workflow file with its own concurrency group. The job would then need no change to `pr-ci.yml` at all, which removes the repo-wide part of this and most of the reason it needs wider sign-off.

petr-sandbox (approved):
Looks good.

PR #450, opened 2026-09-14
perf(tests): build the test schema once and empty it between tests

petr-sandbox:
The backend CI job spends most of its run on DDL rather than on tests. Every test taking `session_factory` built the entire schema and dropped it again, and three other fixtures did the same. Nothing needs a freshly *built* schema — what a test needs is an empty one, and building it from nothing is only one way to get there.

So build it once for the session and empty it with a single `TRUNCATE` between tests. Durability is turned off on the container as well: it is deleted when the run ends, so every fsync it performs is work thrown away, and the per-test reset is almost entirely writes. Neither changes what a transaction sees.

## Where the time was going

Profiling the suite by phase, before this change: setup and teardown of `session_factory` accounted for roughly 85% of wall clock. There is no slow test to find — the slowest single test is under 6s, and a timestamp-gap read of the CI log showed no stall. It is the fixture, several thousand times.

The two levers are independent, measured on one representative test:

| | per test |
|---|---|
| as it was | 0.396s |
| durability off only | 0.191s |
| schema-once + TRUNCATE only | 0.235s |
| both | 0.025s |

## Result

Full suite, same machine, back to back:

| | |
|---|---|
| before | 532.59s |
| after | **82.25s** (6.5×) |

`3131 passed, 6 skipped, 11 deselected, 1 xfailed` either way.

## The one thing to watch

A shared schema means a test that changes DDL hands the change to every test after it. Three in `test_runtime_role.py` do exactly that on purpose — forcing row level security, granting `EXECUTE` back to `PUBLIC` — to watch the boot check refuse it. `rls_harness` now puts both back the way `create_all` leaves them on the way out, beside the default-privileges restore that was already there for the same reason. Without it, six tests fail; with it, none do.

The tests that want a schema of their own rather than merely an empty one — migration parity, the two backfill migrations, the frozen-DDL comparison — build a separate database and are untouched by any of this.

## Test plan

- Full suite before and after on the same machine, twice.
- `ruff format --check`, `ruff check`, `mypy` — clean (243 source files).
- Every directory that looked likely to depend on a freshly built schema, run standalone to catch order or state dependence: `gateway` 283, `bridges` 1696, `clients` 200, `db` 352, `transport` 38, `messages` 19, `provisioning` 9, plus `test_main`, `test_migration_parity`, both backfill migration modules, the frozen-DDL comparison and `test_conftest_tenant_binding` — all pass alone.

## Follow-up, not in this PR

`pytest-xdist -n 4` with a database per worker keyed on `PYTEST_XDIST_WORKER`. Worth measuring once this lands.

Separately: the run that prompted this (20.2 min) was the slowest of the last 25, and the same branch ran 16.2 min an hour earlier against a ~13 min norm — so several minutes of it was hosted-runner noise rather than anything in the tests. The ~13 min floor is the part this addresses.


amaudruz (approved):

petr-sandbox:
CI has now run on this branch.

| | Before | After |
|---|---|---|
| `pytest` step | 19m 28s | **2m 44s** |
| `Backend (core)` job | ~13m | **3m 06s** |

3131 tests pass either way. [Run 34861903433](https://github.com/sandbox-quantum/switch/actions/runs/34861903433).

PR #458, opened 2026-09-14
feat(gateway): bound workspace creation per person (CHOO-2722)

petr-sandbox:
`POST /tenants` is the one route in `gateway/tenants.py` with no tenant to
authorize against — every neighbour checks a membership role, and this one
had nothing between an authenticated person and an unbounded number of
workspaces. Phase 2's design (`docs/old/multi-tenancy-phase2-tenants.md`)
deferred the bound to "what this does not cover"; this builds it and moves
the note into §5 where the route is specified.

It matters more than a row count. `all_tenant_ids()` drives a fan-out per
tenant at boot and a sweep every few seconds, across clients, collaboration
bridges, server connectors and rooms, so self-service creation buys the
deployment steady-state work rather than storage.

## The knob

`GATEWAY_MAX_WORKSPACES_PER_USER`, default `3`, is both the cap and the gate:
at the limit the route is a 403, and `0` closes it outright — how a
deployment that is not ready to offer self-service says so. A negative value
is refused at startup rather than silently behaving like `0`.

Nothing changes for any deployment today: `POST /users` is `require_admin`,
so there is no self-service signup and nobody is near three. The default
exists so the bound is already there when signup opens, rather than being
remembered then.

## Two decisions worth reviewing

**It counts `owner` memberships, not memberships.** Being invited into
someone else's workspace must not spend an allowance, because the invitee
cannot get it back without being removed from a workspace they did not ask
to join.

**Operators are exempt, before anything is read.** `users.role == "admin"` is
the deployment operator bit, and the bypass is what it means everywhere else
in `authz.py`. This bounds self-service; an operator provisioning workspaces
for other people is not that.

## Implementation notes

The count reuses `list_tenant_memberships` rather than asking the database
for a number. No existing exemption returns a caller's *roles* across tenants
— `tenants_of_user` returns ids only — so a numeric answer would have meant a
tenth `SECURITY DEFINER` lookup and a migration, for something bounded to a
handful of rows by construction. (It would also have invalidated the
migration timing rehearsal behind #457.)

`AuthenticatedCaller` gains `is_operator`, read from the `users` row on every
request rather than from the JWT's role claim, so a revoked operator bit does
not stay valid for the life of a cookie. There is a test for exactly that,
using a deliberately stale token.

The bound is checked *before* `create_tenant`, so a refusal cannot leave the
orphan workspace that handler's own docstring warns about — also tested.

## Testing

`3199 passed`. Six new route tests (`TestWorkspaceCreationLimit`) and five
config tests. `just check` and `just typecheck` clean.

Stacked behaviour-wise on nothing; based on `main`. Independent of #457.


petr-sandbox:
1. Simultaneous requests get past the cap (core/switch_core/gateway/tenants.py:295). The route counts the user's owned workspaces, then creates the tenant, then adds the owner row, each in its own session. Nothing locks the count, and the database doesn't enforce the limit. A user at 2 of 3 who sends ten POST /tenants requests at once gets ten new workspaces, because every request sees a count of 2.
   - Fix: take a per-user lock around the count and the insert, e.g. pg_advisory_xact_lock keyed on the user id.
2. Ownership can be handed off to reset the count (tenants.py:252). The cap counts workspaces a user owns now, not ones they created. update_member_role and remove_member only protect the last owner, so user A can create a workspace, promote a second account B to owner, and demote themselves. A's count drops back to 0 and A can repeat forever. B never creates anything, so B's own cap never applies.
   - Fix: count workspaces the user created (e.g. a created_by column on tenants), or also apply the cap when someone is made an owner.

petr-sandbox:
Both addressed in 6f77b19c:
1. Creation now runs under `pg_advisory_xact_lock` keyed on the creator, so concurrent POSTs serialize on the count and insert.
2. The cap reads a new `users.workspaces_created` counter that only goes up (migration `c4e9a1f7b203` seeds it from current ownership), so handing a workspace off no longer resets it.

christian-mcdermott (approved):
approved, just make sure you rebase off main and ensure the migrations wont break

petr-sandbox:
Rebased onto current main. `c4e9a1f7b203` now revises `c4d8e2a1f6b3` (templates visibility) instead of `b8f2d0c41e57`, so main has one Alembic head after merge.

petr-sandbox:
Rebased onto main (f4ada844). `c4e9a1f7b203` now revises main's head `c4d8e2a1f6b3`. Across the whole stack there's a single alembic head (`e3b7c9d2a415`), and the full suite passes.

PR #467, opened 2026-09-15
Show real API key prefix in list_api_keys

### LINKED ISSUE #455: `key_prefix` in the API key list is a prefix of the hash, not of the key
`list_api_keys` in `core/switch_core/gateway/api_keys.py` (lines 33-48) sets `key_prefix=k.key_hash[:12]`.

`key_hash` is `sha256(plaintext).hexdigest()` (line 60), so the value is 12 hex characters of the digest — not of the key.

The gateway UI renders it in a column headed **"Key Prefix"** with an ellipsis appended (`gateway/src/pages/registration-keys/RegistrationKeysPage.tsx`, lines 88-95 and 135-142) and makes it searchable (line 80). An operator holding a key cannot match it against the row: the value is truthfully a hash prefix, but is displayed as though it were the key's opening characters.

### Possible fixes

**A.** Show the real prefix. The plaintext is recoverable — add the existing `config: Annotated[SwitchConfig, Depends(get_config)]` dependency and use `decrypt_token(k.encrypted_key, config.jwt_secret_key)[:12]`, as `reveal_api_key` already does at line 82.

**B.** Rename the field to `key_hash_prefix` and correct the UI header.

A is friendlier but means a list endpoint decrypts secrets on every call, which may not be wanted. Worth deciding explicitly.

### Tests

There is no test file for this router.

amaudruz:
## What

Fixes #455. `list_api_keys` set `key_prefix=k.key_hash[:12]` — 12 hex characters of the sha256 digest, not of the key. The gateway UI renders this in a **"Key Prefix"** column and makes it searchable, so an operator holding a key could not match it against its row.

## Change

Add the existing `config: Annotated[SwitchConfig, Depends(get_config)]` dependency to `list_api_keys` and set `key_prefix=decrypt_token(k.encrypted_key, config.jwt_secret_key)[:12]`, the same way `reveal_api_key` already recovers the plaintext. The column now shows the real opening characters of the key.

Per the issue this is option **A**, chosen deliberately over renaming the field (option B). The tradeoff: the list endpoint now decrypts each of the caller's keys on every call.

The `key_prefix` field name and the UI header "Key Prefix" are now accurate, so no schema or frontend change was needed.

## Notes

There is no test file for this router, so none was updated. Not added here to keep the change minimal — happy to add one if wanted.


amaudruz:
Superseded by #469, which is cut fresh off main with only the api-key fix (da437765) — this branch accidentally carried an unrelated commit (e75f80dc).

PR #487, opened 2026-09-16
feat(telemetry): product usage telemetry for switch-core (CHOO-2806)

christian-mcdermott:
What the Switch core server reports about how it is *used*, as distinct from how
it is behaving — that half is [#491](https://github.com/sandbox-quantum/switch/pull/491),
which this stacks on and shares a wire format with.

Jira: [CHOO-2806](https://sandboxquantum.atlassian.net/browse/CHOO-2806)

**Off unless an operator switches it on.** A Switch server may be a customer's
and the usage may be theirs, so reporting it is a decision someone makes rather
than one they discover. With it off nothing is collected and no request is made.

## What is reported

Three shapes, because the questions want different ones.

**A daily snapshot** carries the counts that describe the installation — users,
rooms, agents, sessions, connectors, messages, and turns. Gathered per tenant
and summed into deployment totals.

**Milestones**, at most once per deployment, each carrying seconds since
install: `deployment_installed`, `first_agent_registered`,
`first_connector_added`, `first_room_created`, `first_room_active`,
`first_session_started`. Together they are the activation funnel.

**Lifecycle events** as things happen — rooms created, archived and deleted,
agents registered and deleted, sessions started and ended, connectors added and
removed, bridges connecting and dropping.

Twenty-two events in all, listed with their triggers in
`docs/old/telemetry-events.md`.

## The rule

**No identifier for anything inside a deployment is ever sent** — not a room,
tenant, agent, user or message; not a name, not an id, not a hash of one. Only
counts, durations, and values from closed sets.

This is enforced rather than documented. `telemetry/catalogue.py` declares every
event and the exact properties it may carry; an undeclared event, an undeclared
property, or a value outside its set raises where the event is built. The test
suite plays the Console's own trick of smuggling a room name into every event
and asserting it never reaches the wire.

The cost is real and worth stating: no per-room or per-tenant breakdowns. We can
say a deployment had 40 active rooms this week, never which.

## Decisions worth reviewing

- **Counted per tenant and summed.** Not because tenants are reported separately
  — they deliberately are not — but because row-level security means a session
  with nothing bound reads nothing, so a single unscoped `COUNT` would answer
  zero on a correctly configured deployment.
- **A session is an agent being present, not a connection.** The heartbeat TTL is
  six seconds and a lapsed stream closes itself, so per-connection reporting made
  a continuously-present agent look like a session every few seconds. Reconnects
  inside a grace window continue the session they resumed.
- **Turns, not senders.** A message classified by who sent the one before it in
  the same room, which is the only way to tell an agent answering a person from
  two agents talking among themselves.
- **No per-message event.** The relay has no batching and no retry because it was
  built for a desktop app emitting a handful of events an hour. Message volume is
  a snapshot count, which supports the same charts.
- **Time to value only for new deployments.** An install date guessed from the
  oldest row would be wrong by an unknown margin in an unknown direction, so a
  deployment that cannot answer honestly reports no milestones at all.
- **One deployment identity across both streams.** `DEPLOYMENT_ID` where an
  operator set one, a generated UUID otherwise — so a deployment reporting both
  operational metrics and product usage is one subject downstream, not two.
- **Two endpoints, deliberately.** `OTLP_ENDPOINT` names a collector, often the
  customer's own. `TELEMETRY_ENDPOINT` names the analytics relay. Same format,
  same client, two destinations.

## Verifying it

`python scripts/otlp_sink.py` is a local stand-in for the relay; point a
development server at it and every event prints decoded as it arrives. The
design note has a per-event checklist with the triggers, and names the two setup
steps without which a third of the list silently cannot fire.

## Review history

Three adversarial review rounds, 29 + 5 + 4 findings fixed. The ones worth
knowing about: every per-tenant query was missing its tenant predicate (counts
multiplied by the tenant count); the activation milestone was wrong four ways at
once; five catalogue events had no emitter; the agent bridge was reporting into a
service that was `None`; and a rejected stream was reporting a session that began
and ended in the same second, which is what filled the first real dashboard.

## Not in scope

Resources, documents, packages, API keys, invitations and room groups are not
instrumented — the commissioned metrics named rooms, agents, sessions,
connectors and usage, and nothing else. Adding them is a catalogue entry and a
call site each.

Nothing here has been verified against Amplitude's own modelling beyond events
arriving; a dashboard built on it should be sanity-checked against the database
once.

[CHOO-2806]: https://sandboxquantum.atlassian.net/browse/CHOO-2806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petr-sandbox (approved):
Makes sense. Did we get the list of metrics from product or will we coordinate with them once this is verified in Amplitude?

christian-mcdermott:
Yes, worked with product to produce a list of metrics

PR #488, opened 2026-09-16
feat(console): give connector events a failure reason and a duration

christian-mcdermott:
## What

Two gaps in the Console's product events, both in the same area:

1. **`connector_installed` had an outcome and no reason.** A failed connector install was a bare count with nothing to act on — the only action event of its kind in the catalogue without a `failure_reason` (compare `agent_cli_action`, `agent_created`, `bridge_connected`).
2. **Nothing in the catalogue carried a duration.** Every event is stamped with wall-clock time, so deltas were computable at the far end, but sends are fire-and-forget with no retry and there is deliberately no correlation id — so pairing broke under concurrency and lost an end whenever a send dropped.

## `TelemetryConnectorFailure`

Named at the point of failure rather than mapped from an error type: these paths only produce a message, and a message cannot be sent.

| code | means |
|---|---|
| `none` | success |
| `unsupported` | connector declared, nothing resolvable to manage it |
| `marketplace_failed` | the marketplace would not register or re-point at its source |
| `install_command_failed` | the host CLI refused the plugin |
| `update_command_failed` | the host's update verb failed |
| `uninstall_command_failed` | the host's remove verb failed |
| `files_write_failed` | the app writes this connector itself, and the write failed |
| `files_remove_failed` | the app writes this connector itself, and removing it failed |
| `files_unimplemented` | the plugin declares `kind: 'files'` and implements no behavior |
| `error` | the operation threw instead of returning a result |

**The reinstall split is the one worth reviewing.** Codex has no update verb, so an update is remove-then-add and `was_reinstall` is true for *both* halves. It cannot tell "nothing changed" from "the agent now has no connector at all" — `uninstall_command_failed` vs `install_command_failed` can, and that is the difference between a no-op and a broken agent.

The codes also separate the two *kinds* of connector, which was the original question behind this change. A marketplace-driven host (Claude Code, Codex) and one whose connector the app writes (OpenCode) fail in entirely different places. `agent_type` implies which today, 1:1 — but that mapping is a dashboard convention that breaks silently the day a second `files` host lands, and the failure codes do not depend on it.

## `duration_ms`

On the three connector events and on `agent_cli_action`, where a CLI install taking four minutes and one taking ten seconds were otherwise the same row.

- **`performance.now()`, not `Date.now()`** — monotonic, so an NTP step or a machine waking mid-install cannot produce a negative duration or an hour that never passed. A negative number in a payload is worse than no number: nothing at the far end can tell it from real data.
- **Whole milliseconds, nothing capped.** A capped duration is a number the operation did not take, which is the invented value the rest of the catalogue refuses. Some of these legitimately include a password prompt somebody left on screen — read them as percentiles, not as a mean. This is documented on the type.
- **Timed around the operation alone**, excluding manager resolution and SSH connect, so the first measurement of a session is not systematically different from the rest.

Nothing free-text, no new identifier: an elapsed time at this resolution names no path, host or command. The disclosure is updated anyway — `console/AGENTS.md` treats widening what is sent as a consent decision rather than a code change. Since #490 that disclosure is `docs/TELEMETRY.md`, behind the link in the dialog, so `duration_ms` and the connector `failure_reason` set are enumerated there.

## Also in here

- `connector_installed` uses `agentTypeOf()` like its siblings instead of an inline `isValidProviderId` ternary.
- `runFiles` folds resolving into its own try on both drivers. That deletes the duplicated guard in `runInstall` and **fixes a real bug**: `update` and `uninstall` on a files connector with no behavior threw out to the UI as a stack trace with nothing reported. Only `install` was guarded.
- Remote's `resolveFiles` carries the connector version like local's, removing three `connectorVersion()` call sites.
- `sentAttributes()` in the telemetry service test read only `stringValue`, so every number and boolean in a payload read back as `undefined` — an assertion that a numeric property was carried would have passed against a payload that dropped it. Now unwraps all three OTLP shapes.

## Testing

New `why a connector operation failed` block pinning each code, including both halves of the reinstall split; `duration.test.ts`; and the existing event assertions updated.

`pnpm run format`, `lint`, `typecheck` and `vitest` are clean on the merged branch — 2645 tests. The three failures and the typecheck errors this description used to list are gone: they were `main`'s, and merging it picked up the fixes.

## Review fixes

A review of this branch found gaps that are now fixed here.

**An operation that threw was reported by nothing.** `resolve()` re-throws a transport failure by design — it is not evidence of absence — so a dead SSH channel mid-install, the most common way a remote install fails, emitted no event at all and reached the renderer as a stack rather than a message. Both drivers now run the operation through `runReportedOperation`; `error` is the code for what it cannot otherwise name, which is why the table above has it.

**A status read that threw took the whole list with it.** The mutating paths were hardened against a `files` plugin with no behavior; the read paths were not, and they fan out over every agent type — so one broken plugin rendered an empty picker instead of one unavailable row. Both fan-outs keep going and say why.

**Codes that did not mean what they said.** A bare catch labelled every `resolveFiles` throw `files_unimplemented`, so a host-side fault was reported as a claim about the shipped app; it has an error class now. A failed *removal* of a files connector was `files_write_failed`, collapsing removal and write failures on the one connector kind these codes exist to separate — hence `files_remove_failed`.

**`TelemetryDurationMs` enforced none of the three properties `AGENTS.md` says it holds.** It is branded now and `startTimer()` is the only thing that can mint one, which immediately caught both controllers typing it as `number`. Relatedly, `expect.any(Number)` matches `NaN` and `Infinity` — values the emitter refuses — so every duration assertion would have passed against a payload that cannot be sent.

**The drivers had already drifted.** Remote read the installed version from the CLI while local read it from the plugin manifest, so the same connector reported a different version per host and fed a different `updateAvailable`. Remote reads the manifest now. Remote `connector_updated` also had no telemetry assertion at all, so the reinstall split was pinned on one of two hand-maintained copies.

Full driver unification — one driver parameterised by the execution context — is the change that removes this class of bug, and is deliberately **not** in here: it is a refactor of its own, not a review fix.

## Not in scope

`catchUpConnectors` calls `switchSetupService.update()`, so the automatic once-per-install catch-up emits `connector_updated` indistinguishably from someone pressing Update. That predates this change, but `duration_ms` makes it worse — background catch-ups will skew the distribution and the failure counts. The catalogue already has the pattern for it (`TelemetryAgentRemoveTrigger`, which exists for exactly this reason); a `trigger: 'user' | 'catch_up'` would sort it. Happy to add it here or separately.

Unrelated to #487, which defines a catalogue for the **core server's** own (not-yet-built) telemetry path. This is Console-side only.


christian-mcdermott:
## Review pass applied — `fcdda9f8`

### Three findings were the same mistake

A comment asserting a property I hadn't verified. Worth naming as a group, because the code was fine and the prose was wrong:

- **`TelemetryDurationMs` said "Nothing is capped".** Every connector command carries `EXEC_TIMEOUT_MS` (120 s), so a hung `plugin install` reports ~120000 and a reinstall-style update stacks two. The docblock now names the timeout wall — *and* the other thing shaping the distribution, which nothing had documented: the connector events time the whole operation, and on a machine that has never registered the marketplace that includes cloning a repo. First install is legitimately much slower than the rest.
- **The providers controller claimed to exclude manager resolution** to avoid the first measurement of a session differing from the rest. `getDependencyManager` is `return localDependencyManager` — it cannot produce that artifact. Comment deleted. The remote one is kept and narrowed, since there resolution may open the SSH connection.
- **`startTimer`'s clamp** was justified by "a future caller passing its own start value" — an API it doesn't have. Monotonic clock, unreachable branch, and its test passed with the clamp deleted. Both gone.

### Real defects

- **`TelemetryConnectorFailure.error` was unreachable.** Nothing mapped onto it and there's no catch-all (unlike `TelemetryCliFailure`, whose `cliFailureReason` defaults to it). Removed rather than faked — and the gap it was pretending to cover, that a *throw* reports nothing at all, is now written down on `connector_installed`.
- **`unsupported` was asymmetric.** A `cli` connector whose host binary can't be resolved was silenced on install by the `attempted` flag while `update`/`uninstall` reported it, so the condition looked like it only happened on update. `install` now uses the same `kind === 'none'` guard as the other two and reports everything past it. `attempted` is gone; new test covers both halves.
- **`console/AGENTS.md` wasn't updated.** It enumerates what a payload may carry, a duration wasn't on the list, and it says widening is a consent decision. `telemetry-copy.ts` names that rule as its authority — so the two halves of the promise disagreed.

### Cleanups

`ConnectorRun` + constructors moved to a leaf module: the remote driver was importing three pure functions from a module ending in `new SwitchSetupService()` over a `LocalExecutionContext`, the exact lesson already recorded on `telemetry/agent-type.ts`. The `files_unimplemented` message no longer differs between drivers. Both duration tests rewritten — the old "does not restart" test passed against a timer that *does* restart, since one sleep's worth satisfies `second >= first`.

### Declined, with reasons

- **Unify the two drivers.** Right in principle, far past this PR. It did catch real drift I'd introduced (the message difference above), which is fixed.
- **Catch throws at the six CLI call sites.** Genuine gap, but pre-existing — none had a `try`/`catch` before this change. Documented on the event rather than silently left.
- **The dead `if (!resolved)` in `runFiles`.** It's required type narrowing: `resolveFiles` returns `null` for a non-files connector, and every caller happens to pre-check. Removing it needs a non-null assertion or a restructure that costs more than the branch.

### Still open (unchanged from the description)

`catchUpConnectors` emits `connector_updated` for the automatic catch-up indistinguishably from a user pressing Update, which now also skews `duration_ms`. A `trigger: 'user' | 'catch_up'` would fix it — happy to add here or separately.

`pnpm run format:check`, `lint`, `typecheck` and `vitest --project node --project main-db` all clean: 3173 passing, one failure (`session-spawner`) that reproduces on untouched `main`.

petr-sandbox:
remote-switch-setup.ts:562 — The new catch in listAgentTypeStatuses reports a failed read as {...unsupportedStatus(agentId), supported: true, refreshError: msg} — but the only consumer (switch-setup/controller.ts listAgentTypeAvailabilityRemote) branches on supported then installed and never reads refreshError. The other consumer, remote-hosts/controller.ts listAgentTypePlugins, has no callers at all. So when the SSH channel dies mid-fan-out, every affected agent type renders as "Its Switch connector is not installed on <host>" with an Install button — a confident false statement about a read that never happened. Before this change the call rejected: ugly, but not false. It's also precisely the "silently degrades to look fine" case CLAUDE.md forbids, and it contradicts the comment two lines above ("a row that could not be read at all says so there rather than vanishing") — it says so in a field nothing reads. The local fan-out at switch-setup-service.ts:334 gets it right via blockedReason; the remote path needs the equivalent.

petr-sandbox (approved):
Just one nitpick.

PR #492, opened 2026-09-16
feat(discord): distributed Discord app 

lbangalosbt:
## Distributed Discord app 

One Discord application **we** register and distribute: a customer clicks **Add to
Server** and never sees a token. The counterpart to the distributed Slack app.
Full design: `docs/old/bridges/DISCORD_DISTRIBUTED_APP.md`.

> Targets `work/multi-tenancy-phase4` (the messaging-install layer it builds on),
> **not `main`**. Draft until 435 merges, then retarget + rebase onto `main`.

## Key decisions — and why

- **One shared Gateway connection, not one per tenant.** Discord issues one bot
  token per app and has no API to create an app per customer, so a single
  connection multiplexes every tenant's guilds and each event is routed to a
  tenant by its `guild_id`. It's owned by the single switch-core pod (already a
  forced singleton), so there's no leader election to arrange.

- **Installs carry no token.** Adding the bot to a guild grants no per-guild
  credential — the bot authenticates with the one deployment-level token. So
  `messaging_installs.encrypted_bot_token` is nullable, a Discord install stores
  nothing, and disconnect revokes nothing. Slack stays strict (its connection
  config still requires a token).

- **Switch enforces all isolation**, because a shared, long-lived, multi-tenant
  socket is the design's central hazard. Every event resolves `guild_id → tenant`
  **fresh** (RLS-scoped re-read), dispatches under `no_tenant()`, and binds
  per-room — no tenant is ever cached on the connection. Four guards, each with a
  test: **G1** per-event scoping · **G2** identity keyed by (tenant, Discord user
  id) · **G3** unknown guild dropped (fails closed) · **G4** DMs dropped, no DM
  intent.

- **Mention-only by default.** `message_content` and `members` are privileged
  intents that close the connection past ~100 guilds unless verified, so both are
  gated behind flags, default off.

- **The per-guild bridge is a normal, inert `CollaborationAdapter`.** Under
  `event_delivery: shared` it opens no socket of its own and routes through the
  shared connection — reusing all existing per-guild logic (posting, roles,
  slash, room map) unchanged, so `get_adapter`, moderation and the operator list
  keep working.

## Shape

Ordered, individually reviewable commits: refactor (extract `DiscordConnection`)
→ **1** tokenless install → **2** config + installer → **3** inert bridge → **4**
shared connection + per-event routing + global slash + removal + guards.

## Testing

Full suite green (**3472 unit + 11 integration**). **Validated live against real
Discord**: install → tokenless record + inert bridge; `@mention` → agent reply;
two guilds in two tenants over the one connection with **zero cross-talk** (same
Discord user → two tenant-scoped puppets); DMs not delivered; Slack untouched.

## Follow-ups (out of scope)

- **Eager attach:** an inert bridge provisions agent roles at start before its
  first message injects the shared connection — fails loud and self-heals; worth
  attaching eagerly.
- **Phase 2** (mostly *shared* code): self-serve install (installing is
  admin-gated today), callback → redirect, and the operator list showing a
  shared-connection install's status/revoke.


core/switch_core/bridges/collaboration/discord/install.py
```diff
@@ -0,0 +1,206 @@
...
+#: commands as native slash commands. Pinned in DISCORD_DISTRIBUTED_APP.md and
+#: compared by a test.
+SCOPES: tuple[str, ...] = ("bot", "applications.commands")
+
+#: The least-privilege guild permissions the authorize URL requests, as named
+#: bits summed into the bitfield Discord takes. Each is a capability the adapter
+#: actually exercises — nothing here is speculative, and adding one is a
+#: decision this table records. The permission table in DISCORD_DISTRIBUTED_APP.md
+#: lists the same bits, and a test pins the resulting integer.
+_PERMISSION_BITS: dict[str, int] = {
```
christian-mcdermott, line 59: do we need `"create_public_threads": 1 << 35` ?

deploy/remote/helm/switch/templates/_helpers.tpl
christian-mcdermott, line 611: `DISCORD_APP_*` is documented in `.env.example` but wired nowhere - the Helm chart and both docker-compose files pass only `SLACK_APP_*`, so the feature cannot be enabled on any supported deployment path. Is this intended?

core/switch_core/bridges/collaboration/discord/adapter.py
```diff
@@ -159,10 +198,23 @@ def __init__(self, *, config: DiscordConnectionConfig) -> None:
...
+        # socket is per bot token and the adapter is per guild. Intents are
+        # built here and handed over, so the socket owner does not decide them.
+        #
+        # A self-registered bridge owns its connection, built now from its
+        # token. A distributed (shared-delivery) bridge owns none: it is inert
+        # until it is attached to the one deployment-level connection, which is
+        # not built yet — so `_connection` stays None and every outbound path
+        # fails loud through `_require_connection` rather than pretending.
+        self._connection: DiscordConnection | None = None
+        if config.event_delivery == "own_connection":
```
christian-mcdermott, line 211: Record ownership at construction rather than inferring it:
```
self._connection: DiscordConnection | None = None
self._owns_connection = config.event_delivery == "own_connection"
if self._owns_connection:
    assert config.bot_token is not None
    self._connection = DiscordConnection(...)

async def stop(self) -> None:
    if self._owns_connection and self._connection is not None:
        await self._connection.close()
    else:
        self._connection = None      # detach, don't close
    self._webhooks.clear()
```
Detaching (rather than leaving the reference) means a restarted bridge re-attaches on its next event instead of holding a possibly-replaced connection. Add a test that attaches, stops, and asserts the shared client is still live.

core/switch_core/bridges/collaboration/install_routes.py
christian-mcdermott, line 296: Should platform be a type rather than a string?

christian-mcdermott, line 202: curl -X POST https://<MESSAGING_PUBLIC_URL>/messaging/discord/events -d '{}' → repeatable 500 + traceback, from anyone. Not a data or authz problem (verification runs first, no attacker body is parsed) — it's unbounded log/alert noise plus a "retry me" status for a permanent condition, exactly what _inbound's docstring says the codes are chosen to avoid.

Fix. Let the installer declare its webhook surface and 404 the rest, matching the existing posture ("not found rather than an explanation: the caller is unauthenticated and learns only that there is nothing here"):

```
# MessagingAppInstaller
WEBHOOK_ENDPOINTS: ClassVar[frozenset[WebhookEndpoint]] = frozenset(
    {"events", "interactive", "commands"}
)
# DiscordAppInstaller
WEBHOOK_ENDPOINTS: ClassVar[frozenset[WebhookEndpoint]] = frozenset()

In _inbound, before authenticate:

installer = service.installer_for(platform)          # None → existing 404 path
if endpoint not in installer.WEBHOOK_ENDPOINTS:
    return Response(status_code=404)
```

core/switch_core/bridges/collaboration/discord/adapter.py
```diff
@@ -211,123 +263,103 @@ async def start(
...
-        visible in the logs rather than silent.
+    async def stop(self) -> None:
+        # A shared-delivery bridge has no connection of its own to close;
+        # stopping it is just dropping its per-guild state.
+        if self._connection is not None:
+            await self._connection.close()
+        self._webhooks.clear()
+        logger.info("Discord adapter stopped")
+
+    def ensure_shared_connection(self, connection: DiscordConnection) -> None:
```
christian-mcdermott, line 312: `lifecycle.start()` → `BridgeCore.start()` → `_create_agent_identities()` runs at bridge start, which on a distributed install is always before any inbound message. Each `create_agent_identity` hits `_get_guild` → `_require_client` → `_require_connection`, which raises "this Discord bridge uses shared delivery and is not attached to the shared Gateway connection yet". `_create_agent_identities` catches per agent and logs, ending with "Provisioned discord identities for 0 of N agents; N failed".

`ensure_shared_connection` only assigns the field — it triggers no re-provision — and nothing else re-runs identity creation. So the per-agent guild roles that make @agentname autocomplete in the Discord composer are never created, permanently, across restarts.

Worth being precise: this is not a sticky disable. `_disable_agent_roles` is only called on discord.Forbidden and the 250-role limit, so `_agent_roles_off_reason` stays None. The work is simply never attempted again.

In the same window, every `send_message` / `send_attachment` / `update_message` raises, so a Switch room linked to a Discord channel cannot be posted into until a human types in Discord. The PR describes this as "fails loud and self-heals"; it fails loud and stays failed.

Fix — preferred. Attach eagerly at boot rather than lazily on first event, which removes the "first message is special" asymmetry that produces this. In DiscordGatewayClient.start(), after the socket is ready, walk the active Discord installs and attach each adapter. That requires collab_lifecycle.start_all() to run first, so invert the two in main.py (bridges up, then the gateway attaches to them) — safe, because an inert bridge does nothing until attached anyway.

Fix — smaller. Make attach do the work it unblocks:
```
async def attach_shared_connection(self, connection: DiscordConnection) -> None:
    if self._connection is not None:
        return
    self._connection = connection
    if self._on_attached is not None:
        await self._on_attached()     # BridgeCore re-runs _create_agent_identities
```
BridgeCore.start() registers _on_attached for shared bridges. _create_agent_identities is already close to idempotent — create_agent_identity early-returns on agent_name.casefold() in self._agent_role_ids and adopts a same-named existing role rather than duplicating.

core/switch_core/main.py
```diff
@@ -593,6 +605,31 @@ async def lifespan(app: object) -> AsyncIterator[None]:
...
+        # not None here. Asserted rather than branched to say that out loud.
+        assert install_service is not None
+        discord_gateway = DiscordGatewayClient(
+            bot_token=config.discord_app_bot_token,
+            message_content=config.discord_app_message_content,
+            members=config.discord_app_members,
+            install_service=install_service,
+        )
+        try:
+            await discord_gateway.start()
```
christian-mcdermott, line 626: A failed gateway start never reconnects. discord.py's own reconnect logic lives inside client.connect(), which for two of DiscordConnection.connect's three failure modes never gets established (login raising on a bad token; the connect task completing with an exception), and for the third — the 30s readiness timeout — connect calls self.close() first. So a start failure is terminal for the life of the process while the log promises recovery.

Fix: A supervised retry with backoff, its task handle kept for shutdown:
```
async def start_with_retry(self) -> None:
    delay = 5.0
    while True:
        try:
            await self.start()
            return
        except Exception:
            logger.exception("Discord gateway start failed; retrying in %.0fs", delay)
            await asyncio.sleep(delay)
            delay = min(delay * 2, 300.0)
```

core/switch_core/bridges/collaboration/install_service.py
```diff
@@ -458,18 +465,31 @@ def revocation(self, *, platform: str, event: InboundWebhook) -> Revocation | No
...
     async def resolve(self, *, platform: str, event: InboundWebhook) -> WebhookTarget:
-        """Turn a workspace id into the one bridge entitled to the event.
+        """Turn a webhook event's workspace into the bridge entitled to it."""
+        installer = self._installers.get(platform)
+        workspace_id = installer.workspace_of_event(event.payload)
+        return await self.resolve_by_workspace(
+            platform=platform, workspace_id=workspace_id
+        )
+
+    async def resolve_by_workspace(
```
christian-mcdermott, line 475: `resolve_by_workspace` does two independent sessions per event: tenant_of_messaging_install(...) (the exempt / SECURITY DEFINER lookup) and then store.get_for_workspace(...) under a tenant_session. Both run before dispatch_inbound → _handle_message, where the cheap drops live — own-bot posts, the bridge's own webhook echoes, non-default/reply message types, dedupe.

Fix - Two parts:

Pre-filter before resolution — these need no tenant and no DB:
```
if message.author.id == self._connection.bot_user_id:
    return
if message.type not in ALLOWED_MESSAGE_TYPES:
    return
```
Promote `_ALLOWED_MESSAGE_TYPES` from `DiscordAdapter` to a module constant so the gateway isn't reaching into a private. Webhook-echo dropping can't move (_webhook_ids is per adapter), but it's the smaller share.

Cache guild → target. A `dict[str, WebhookTarget]` on the gateway client, invalidated explicitly from `install_service` on install / disconnect / revoke. That keeps the deliberate scoped re-read on the cache-miss path (where the isolation guard matters) and removes it from the steady state.

gateway/src/pages/collaborations/DisconnectAppDialog.tsx
christian-mcdermott, line 66: For Discord all three claims are false: encrypted_bot_token is NULL (the bot uses the deployment-level application token), events arrive over the shared Gateway socket, and disconnect skips revocation entirely. An operator disconnecting a Discord install is told a credential was revoked when the bot is still authenticated in that guild until someone removes it in Discord.

Fix. Per-platform copy, keyed off platform (client-side, or served alongside the existing capability flags):
```
const INSTALL_COPY: Record<string, { grant: string; delivery: string; removal: string }> = {
  slack: {
    grant: "a token that Switch holds and you cannot see or rotate",
    delivery: "over HTTPS",
    removal: "revokes that token at the platform first",
  },
  discord: {
    grant: "no per-install token — the bot authenticates with this deployment's own application token",
    delivery: "over the shared Gateway socket",
    removal: "stops delivery to this server; the bot stays authenticated until it is removed from the server in Discord",
  },
}
```
The dialog matters most — telling someone a credential was revoked when it wasn't is the kind of thing that surfaces in an incident review.

christian-mcdermott (commented):
Looks good, I just found some things that may be worth addressing first.

deploy/remote/helm/switch/templates/_helpers.tpl
lbangalosbt: missed pushing change!! done.

core/switch_core/bridges/collaboration/install_routes.py
lbangalosbt: will add as a quick follow up

lbangalosbt: good catch! added

core/switch_core/bridges/collaboration/install_service.py
lbangalosbt: Did the pre-filter. will address cache as a follow up.

core/switch_core/bridges/collaboration/discord/adapter.py
lbangalosbt: Thanks for clarifying this issue. The fix is a mix of both. The flow is now: bridges start immediately (inert if shared, waiting to be handed the shared connection), the gateway connects on its own schedule with retry, and attach re-runs provisioning whenever it lands.

PR #505, opened 2026-09-18
(CHOO-2805) Add a Stop button to agent status messages on all chat platforms

simonflk-sb:
## Purpose

Stopping a working agent from chat used to mean knowing and typing a command — `!interrupt`, or `/interrupt @agent-name`. This branch puts a **Stop current work** button directly on the message that shows what the agent is doing, on all five supported chat platforms: Slack, Mattermost, Discord, Telegram and Teams.

It also fixes the thing that made that message untrustworthy. A status card is redrawn on every tool call, and when a platform refused a redraw the card simply froze — still showing an agent at work, sometimes hours after the turn had finished, with nothing anywhere saying why. The card now either keeps up, or says out loud that it has stopped.

## Key Changes

**Stopping an agent from the message itself.** A running turn's status message carries a Stop button while there is something to stop. Pressing it asks the agent to end the turn it is running now; the agent decides how quickly it can, and the same message reports when it has. The button disappears when the turn ends and the **Show activity** fold takes its place.

**A press can only stop what the presser was looking at.** The control carries the id of the turn it was drawn for. A press from a stale view — a scrolled-back card, or a client that has not caught up since a new turn started — is refused rather than silently stopping whatever is running now. A message queued behind a running turn gets its own card whose button stops the work *in front* of it, and the card says so in as many words.

**Presses are authorised and answered privately.** Whoever presses goes through the same permission check as a typed interrupt. The outcome, accepted or refused, is shown to that person alone — nobody else in the channel sees that you stopped something, or that you were not allowed to.

**A press is answered in the thread it was made in.** Presses on an attention reply now resolve to the turn they are actually about, and the answer goes back into the thread the press came from rather than the thread the card was originally posted in.

**Slack streams survive a restart.** A streaming card left open by a restart used to be retried forever. It is now adopted and closed, or given up on deliberately, and the turn stays owed until one of those actually happens.

**A card Slack will never accept again says so.** Slack can put a message into a state where it refuses every call against it — update, append, stop and delete alike — which makes the card unrepairable. Switch now stops asking, and posts one short message beside the card explaining that it is frozen and that the turn itself was unaffected. That notice goes into the same thread as the card, is sent once, and is skipped entirely for a card more than an hour old, where it would reach nobody and only push the channel along.

**Rate limits say what they stopped.** Slack and Telegram both hold back work after a 429, and both used to do it silently — every card freezing at once with nothing in the log to explain it. Both now report the pause and its lift, naming the platform, what is held, and the consequence.

**Setup docs** for Discord, Mattermost, Teams and Telegram describe the button, what each platform requires for it, and how it degrades where a platform cannot carry it.

## Design Decisions

**The control names its turn rather than trusting the card's position.** Binding the press to a turn id is what makes a stale press refusable. The alternative — stopping whatever is currently running — is the behaviour most likely to destroy work the presser never intended to touch.

**Where a platform cannot carry the button, it is not drawn at all.** Discord caps a component id at 100 characters and Telegram at 64 bytes. A provider whose turn ids exceed roughly 90 and 60 respectively gets no button and a warning in the log, rather than a control that errors when pressed. The typed interrupt still works in every one of those cases.

**A frozen card and a reader who has been told about it are tracked as two separate durable facts.** The wedge is recorded the moment the platform refuses the edit, and it is what stops the card being probed again after a restart. Whether the explanation actually reached the channel is recorded separately, only once it has an outcome — delivered, or deliberately skipped as too old. Holding both in one flag loses whichever is written second: record at discovery and a notice the platform refused is never retried; record on delivery and every restart goes back to probing a card that can never answer.

**The wedge is keyed on the message reference, not on the turn.** A card can be reposted under the same journal row. A flag meaning only "something here froze once" would silence the healthy replacement permanently, which is a worse and quieter fault than the stale card it was added to prevent.

**A refused notice keeps the turn undrawn.** `send_message` reports a refusal by returning nothing rather than by raising, and the adapter reports a given-up card only once. So the turn staying owed is the only thing that produces a second attempt — including for a turn that ends on the same pass the freeze is discovered, which otherwise writes its completion receipt and never publishes again.

**Slack and Telegram share one cooldown implementation**, which names its own scope. Telegram's hold is bot-wide, not per-chat: Telegram meters both a single chat and the bot across all of them, and a 429 does not say which limit was hit, so there is no sound basis for scoping the hold more narrowly.

## Technical Notes

- `RichContentWedged` is a distinct adapter failure, separate from an ordinary redraw failure, because an ordinary one is worth retrying and this one never is. It is raised on the attempt that discovers the state; later attempts on the same message return quietly.
- The activity journal row gains two keys, `wedged` and `wedge_notice`, both holding a message reference. Nothing in the wild carries an older shape, so there is no migration.
- Slack's registry of given-up messages no longer shares the 100-entry bound that limits open streams; eviction there resumes asking about a message that cannot answer.
- Teams cards declare Adaptive Card schema 1.5 only while a turn is running, since that is what `Action.Execute` requires. A card with nothing to stop is unchanged, so older clients are unaffected outside a live turn.
- Mattermost needs one extra API call to preserve props the server itself set. It is made only when a redraw actually changes the buttons — roughly twice per turn rather than once per redraw — though a freshly restarted bridge pays it on its first redraw of each post.

## Release Notes

> [!IMPORTANT]
> **Mattermost buttons need configuration.** Set `callback_base_url` to the base URL (scheme, host and port, no path) that the Mattermost server can reach switch-core's callback listener on. Left unset, Mattermost posts carry no buttons and the bridge logs one warning at startup; `!interrupt` continues to work.

> [!IMPORTANT]
> **Discord buttons need a publication webhook this application owns.** A status published through a webhook owned by something else carries no components, and the typed interrupt is then the only way to stop an agent.

> [!NOTE]
> No database migration. No schema change, and the two new journal keys are additive.

> [!NOTE]
> Telegram draws the button in its red `danger` style, which arrived in Bot API 9.4 and is only rendered by clients released after February 2026. Older clients show an ordinary button and no error, so the label carries the warning on its own.

## Testing

- Full collaboration and sessions suites: **2570 passing**. `mypy` clean across 291 source files; `ruff format` and `ruff check` clean.
- Coverage added for the interrupt control on each of the five platforms, for the turn-binding refusal, for the queued-message case, and for the Slack streaming, restart and rate-limit paths.
- The frozen-card behaviour is covered for discovery, restart, a notice the platform refuses, a card too old to be worth explaining, a replacement card posted under the same journal row, and a turn that ends on the same pass the freeze is found. Each of those fixes was confirmed by breaking it deliberately and checking the test failed before restoring it.

> [!WARNING]
> **The frozen-card notice has never been seen against real Slack.** Every test of it runs against a fake adapter, so nothing here demonstrates Slack rendering that message or placing it correctly. Producing it on demand needs a card to freeze mid-stream, which is not reproducible to order. Existing frozen cards in a workspace are well over an hour old and will correctly be left alone.

Still unverified by hand: **Show activity** and **Stop** on a Discord or Mattermost attention message, and the Console link on Slack, Teams and Telegram.


simonflk-sb:
merged in #507

